PR #1, opened 2016-03-04
Pom fixes for [JENKINS-33188]

svanoort:
Fixes POM dependencies, adapt to new parent pom, improve plugin description and explicitly specify it (resolves bug with descriptions being duplicated).


svanoort:
@reviewbybees

pom.xml
```diff
@@ -46,8 +46,9 @@
     </scm>
 
     <properties>
-        <jenkins.version>1.609.3</jenkins.version> <!-- but see Jenkins-Version override below -->
+        <jenkins.version>1.609.3</jenkins.version>
         <workflow.version>1.14</workflow.version>
+        <jenkins-test-harness.version>2.1</jenkins-test-harness.version>
```
jglick: That is the default for this POM version, not necessary to override it unless you want to go newer.

jglick:
:bee:

PR #2, opened 2016-03-11
[Fix JENKINS-33486] Replace FlowNodeNavigator with FlowGraphWalker

tfennelly:
@jenkinsci/code-reviewers and @reviewbybees


rtyler:
2rmoh

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -299,15 +300,12 @@ public static FlowNode getFlowEndNode(FlowExecution execution) {
...
-        FlowNodeNavigator nodeNavigator = new FlowNodeNavigator(new FlowNodeNavigationListener() {
-            @Override
-            public void onNode(FlowNode navNode) {
-                if (navNode.getParents().contains(parentNode) && !nodes.contains(navNode)) {
-                    nodes.add(navNode);
-                }
+        FlowGraphWalker walker = new FlowGraphWalker(parentNode.getExecution());
+        for (FlowNode node : walker) {
+            if (node.getParents().contains(parentNode) && !nodes.contains(node)) {
+                nodes.add(node);
```
jglick: BTW using a `LinkedHashSet` would simplify this a bit:

```java
if (node.getParents().contains(parentNode)) {
    nodes.add(node);
}
```

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -347,16 +345,13 @@ public void onNode(FlowNode navNode) {
...
-        FlowNodeNavigator nodeNavigator = new FlowNodeNavigator(new FlowNodeNavigationListener() {
-            @Override
-            public void onNode(FlowNode node) {
-                if (!unsortedNodes.contains(node)) {
-                    unsortedNodes.add(node);
-                }
+        FlowGraphWalker walker = new FlowGraphWalker(nodeList.get(0).getExecution());
+        for (FlowNode node : walker) {
+            if (!unsortedNodes.contains(node)) {
+                unsortedNodes.add(node);
```
jglick: ditto

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
```diff
@@ -211,25 +210,22 @@ public static RunExt create(WorkflowRun run) {
...
-                        // the end time for the run.
-                        runExt.setEndTimeMillis(nodeTime);
-                    }
-
-                    if (StageNodeExt.isStageNode(node)) {
-                        StageNodeExt stage = StageNodeExt.create(node);
-                        runExt.addStage(stage);
-                        runExt.setPauseDurationMillis(runExt.getPauseDurationMillis() + stage.getPauseDurationMillis());
-                    }
+            FlowGraphWalker walker = new FlowGraphWalker(run.getExecution());
```
jglick: or

```java
FlowGraphWalker walker = new FlowGraphWalker(execution);
```

jglick:
:bee:

svanoort:
@tfennelly I concur with all Jesse's remarks here.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
tfennelly: @jglick Sorry ... I don't see how it's simplified ... looks like the same code. Only diff I can see is `List` Vs `Set`.

The `HashSet` might be a bit more optimised, but I wasn't going to assume anything wrt hashing/equals impls + I would think that the lists are so small that the performance diff would be near zero. Maybe those assumptions are not right though.

svanoort:
:bee:  and merging, I'll do the hashset optimizations and submit as a separate PR

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: > looks like the same code

No need to call `contains`.

jglick: > I wasn't going to assume anything wrt hashing/equals impls

[the impls](https://github.com/jenkinsci/workflow-plugin/blob/82e7defa37c05c5f004f1ba01c93df61ea7868a5/api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java#L315-L327)

tfennelly: @jglick How about making those methods `final` ? `HasSet` etc are great until someone reimpls those methods and barfs the whole thing up.

jglick: Huh? This is a local variable.

svanoort: @tfennelly / @jglick Worth mentioning that one of the optimizations I tested was looking to see if HashSet was better for this check.  It's actually significantly worse for performance, because in almost all cases, the number of parents is 1, and creating a hashset adds an overhead vs. iterating an existing list.  The only cases where parents will be > 1 are where there's a parallel set block, and the size of the Set is still too small to merit it.

PR #4, opened 2016-03-12
[JENKINS-33624] Improve Perfomance with Large Pipelines

svanoort:
[\[JENKINS-33624\]](https://issues.jenkins-ci.org/browse/JENKINS-33624) Improves performance when working with pipelines containing many nodes
- Cap number of child nodes returned in a StageFlowNodeExt to 100, because multi-megabyte objects were being returned.  Performance on requests went from 10 seconds to <50ms. 
- Nodes still have to be trivially touched to check for failures, but not creating API response objects & not doing real parsing of the flownodes saves tremendous amounts of time.

**TODO:**
- ~~[ ] Cache StageViewExts when they are not part of a running build (limit to a few hundred entries & time-limit them - they should be fairly small objects, and caching will improve performance of the expensive describe calls)~~ - Better approach here
- [x] Smarter caching: create a fixed-size (?) cache for Run metadata ( everything in RunExt )  when run is complete - locally only, not pushed (investigating some performance regressions here)
- [x] Finding the node a pipeline step ran on was generating a StackOverflow - replace recursive impl with queue-based method
- [x] Replace the Action-based approach to caching sorted FlowNode lists with caching impl
- [x] Calculate the execution time of a stage more efficiently
- [x] See why profiling indicates that getAction is a bottleneck - profiling quirk?
- [x] ~~Linear search is used for finding the index in sorted list of a StageViewNode.  Use binary search to improve scaling.~~  - Appears to hurt, not help performance.
- [x] Find out why caching the RunExt wasn't helping - answer: we needed to hit caches in places it was missed.
- [x] Use ItemListener.onLocationChanged to handle job renames with caching (and risk of a double-rename causing bad data)
- [x] Optimize initial load of runs data - use a HashMap based algorithm to map info to stage nodes with a single scan (or worst case, two scans
- [x] Create RunExt using the fast flow analyzer
- ~~[ ] Replace existing RunExt methods with the fast scanner.~~ - FlowAnalyzer splitting to separate PR so we can release existing fixes. 
- [x] Create cache for StageExt objects with child nodes realized, separate from the RunExt entries (reduces data transfer while still allowing caching)
  - RunExt can still partially cache, we just need to do a scan of the stage for children and create a clone of the StageExt in the second cache
- ~~[ ] Look at using a loader implementation for the RunExt cache, so that multiple threads don't try to do the computation? - CacheLoader.load (need to check it's threadsafe though).~~
  - No, it's risky to block on the initial computation since it is rather long, for now.
   
- [x] Verify that the UI can still get full data for a stage run when needed - tree view.
- ~~[ ] Pull out the List<FlowNode> caching - generate it once and pass it into the method that need it as arguments (only needed in Stage parsing)~~ - replaced by the new FlowAnalyzer algorithm
- ~~[ ] Try dropping in Caffeine cache as a replacement for guava cache, since benchmarks say it is [3-10x faster](https://github.com/ben-manes/caffeine/wiki/Benchmarks) and the API is the same~~ - largely irrelevant to performance, caching is super cheap, also only supports Java 8.
~~- [ ] Enable the cache stats to be visible by API~~
- [x] Look more closely at cache settings & performance.  They may not be sensible.  
   - Seem strong enough, especially since the executionCache and execNodeNameCache will go away soon with the new flowanalyzer (just caching the RunExt).
- [x] Optional: fix URL handling to allow isolated unit test & benchmarks of *Ext classes
# Splitting
- [x] Once above done, create commit removing FlowAnalyzer, and a PR that reverts that commit for additional development on flow analyzer


rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -44,33 +62,70 @@
...
     private FlowNodeUtil() {
     }
 
-    public static long getNodeExecDuration(FlowNode node) {
-        List<FlowNode> childNodes = getChildNodes(node);
-        if (!childNodes.isEmpty()) {
-            long startTime = TimingAction.getStartTime(node);
-            long endTime = TimingAction.getStartTime(childNodes.get(childNodes.size() - 1));
+    public static boolean isNotPartOfRunningBuild(FlowExecution execution) {
+        Queue.Executable executable = null;
```
svanoort: @jglick  I do *not* consider this overall ready for review, but wonder if this test makes sense?  It's borrowed from existing logic, but I'm not 100% sure if it is correct (I'm seeing some odd behaviors).  Is there a better way?  Perhaps something like the following?

```java
return (execution instanceof CpsFlowExecution && ((CpsFlowExecution)execution).isComplete());
```

tfennelly:
> Cap number of child nodes returned in a `StageFlowNodeExt` to 100.

I was thinking about this the other day. Ultimately, I think this needs a new server-side API to replace some of the client side code (to process and consolidate/mashup the data returned to the client, vs the client getting everything and doing the mashup itself) and/or to get rid of the API that returns all nodes. Hindsight is 20/20 vision, but when this was originally written there was obviously a misunderstanding (on my behalf) about how this flow graph would grow. I obviously had it in my head that it would never get so huge. Grrrr :anguished:

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Based on conversation, "isComplete" was the recommended method.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
```diff
@@ -232,6 +232,7 @@ public static RunExt create(WorkflowRun run) {
 
                 if (StageNodeExt.isStageNode(node)) {
                     StageNodeExt stage = StageNodeExt.create(node);
+                    stage.addStageFlowNodes(node);
```
svanoort: This is needed currently since the StageExt serves itself from the appropriate RunExt cache, if present. 

THE CATCH is that this inflates the size of the RunExt object serialized and sent to clients.  

The UI could directly consume the fully-realized RunExt with full nested child nodes (and avoid calling requests to the FlowNodeAPI Describe API for each stage start node).

Until that happens, a way to avoid bloat on the response but still keep caching simple:
* Create a RunExtTrimmer class that wraps a RunExt, but removes the child nodes from StageExt objects returned (and saves them). 

Catch: you still have to do full stage processing up from on the RunExt.  Perhaps we should avoid that in cases where the Run is still building since it's expensive to do this.  But in other cases, you *will* have to do it at some point, might as well be up front.

svanoort:
@reviewbybees

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeListCacheAction.java
```diff
@@ -1,81 +0,0 @@
...
-import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
-/**
- * Simple action for caching pre processed/filtered/sorted node lists.
- *
- * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
- */
-public class FlowNodeListCacheAction extends InvisibleAction {
```
jglick: Did you check what happens to historical build records that included this? I would expect that _Manage Old Data_ becomes activated, though what it shows in the case of flow node XML files, and what happens when you click the _Discard_ button, I am not sure.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -44,33 +61,88 @@
  */
 public class FlowNodeUtil {
 
+    protected static final Cache<String,List<FlowNode>> executionCache = CacheBuilder.newBuilder().maximumSize(100).build();
```
jglick: :ant: `protected static` makes no sense on an uninstantiable (so, effectively `final`) class. You can just delete `protected` and it will still be accessible from within the package.

jglick: :bug: a `FlowNode` will in general hold a strong reference to a `WorkflowRun` and thus also `WorkflowJob`, which is not something you want in a `static` cache.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -44,33 +61,88 @@
...
+
+
+    // Larger cache of run data, for completed runs, keyed by flowexecution url, useful for serving info
+    // Actually can be used to serve Stage data too
+    // Because the RunExt caps the total elements returned, and this is fully realized, this is the fastest way
+    public static final Cache<String, RunExt> runData = CacheBuilder.newBuilder().maximumSize(1000).build();
```
jglick: :ant: `public` on a mutable object? Really?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -44,33 +61,88 @@
...
+
+    static final Cache<FlowNode,String> execNodeNameCache = CacheBuilder.newBuilder().weakKeys().expireAfterAccess(1, TimeUnit.HOURS).build();
```
jglick: Also :bug: as a rule you should not use `static` fields for caches in Jenkins. It is too easy to wind up with state carrying over between sessions, leading to mysterious problems in functional tests. Use an instance field in an `Extension` instead, to ensure that the cache is properly scoped to a `Jenkins` instance.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -44,33 +61,88 @@
...
+            try {
+                runData.put(exec.getUrl(), run);
+            } catch (IOException ioe) {
+                LOGGER.severe("Can't get execution url for execution, IOException!");
             }
+        }
+    }
+
+    /** The nuclear option: clear all caches */
+    public static void invalidateAllCaches() {
```
jglick: :ant: having such a method is an antipattern. Define your caches with `Jenkins` scope to begin with and this is not necessary.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: There are only a couple dozen calls to `FlowNodeUtil` static methods from outside the class. A simple search-and-replace of `FlowNodeUtil.findSomething()` → `FlowNodeUtil.get().findSomething()` suffices to fix the static cache mistake. Later, if desired, you can start to thread the `FlowNodeUtil` singleton instance through different methods to reduce the number of `get()` calls, ultimately using `@Inject FlowNodeUtil util` in `Extension`s to avoid the call entirely.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @jglick The latter is addressed in the latest commit.  The former is probably going to change shortly anyway.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @jglick It exists as an escape valve in the event of very unforeseen circumstances, not just for tests.  The intent is that ultimately it can be invoked by users by API in the rare event that someone manages to generate a situation that abuses the caches.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -44,33 +61,87 @@
  */
 public class FlowNodeUtil {
 
+    protected static final Cache<String,List<FlowNode>> executionCache = CacheBuilder.newBuilder().weakValues().maximumSize(100).build();
```
jglick: Probably a :bug: since there is not likely anyone holding this `List` strongly, so the cache will get discarded on every GC cycle.

When I said you should avoid holding `FlowNode`s strongly in a cache, what I meant is that you should change the type of the cache, to their `String` IDs for example.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: Anyone who gets into such a situation (and knows it, and is in direct consultation with you) can trivially clear any fields, even `private`, using `/script`.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeListCacheAction.java
svanoort: @jglick I had tested with it and didn't get any notification; after rechecking the manage old data area, it shows nothing. 

Test process:

- Launch jenkins via docker image
- Install pipeline stage view, and restart after install (v1.0)
- Create a simple pipeline with stages and echo statements, and run it
- Refresh page a few times to ensure it gets hit and cached by pipeline stage view 
- Upload HPI for new pipeline stage view Rest API + UI, restart jenkins
- Look at 'manage old data'  -- "No old data was found."

Unless there is a necessary step here that is missing in testing, it looks like it just ignores the unrecognized action.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @jglick Addressed in latest commit by using weak values

svanoort: @jglick Ah, okay it did update with latest commit.  In any case, the executionCache only needs to live as long as the current run is being actively analyzed and no longer.  One alternative would be to use a very small cache size but  strong values, assuming holding references to a few WorkflowJobs is not the end of the world.

jglick: Never hold strong references to `Item`s.

svanoort: @jglick Could you explain why that is so?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Removed.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @jglick The caches are dealt with now.  A little awkward, but it works, and it will be simpler when I can fully remove the extra caches.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/UndefinedWorkflowTest.java
```diff
@@ -24,6 +24,7 @@
 package com.cloudbees.workflow.rest.endpoints;
 
 import com.cloudbees.workflow.Util;
+import com.cloudbees.workflow.flownode.FlowNodeUtil;
```
kzantow: unused?

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/ParallelStepTest.java
```diff
@@ -23,6 +23,7 @@
  */
 package com.cloudbees.workflow.rest.endpoints;
 
+import com.cloudbees.workflow.flownode.FlowNodeUtil;
```
kzantow: unused?

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -23,6 +23,7 @@
  */
 package com.cloudbees.workflow.rest.endpoints;
 
+import com.cloudbees.workflow.flownode.FlowNodeUtil;
```
kzantow: unused / whitespace changes only here?

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/FlowNodeAPITest.java
```diff
@@ -24,6 +24,7 @@
 package com.cloudbees.workflow.rest.endpoints;
 
 import com.cloudbees.workflow.Util;
+import com.cloudbees.workflow.flownode.FlowNodeUtil;
```
kzantow: unused?

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/InputStepTest.java
```diff
@@ -24,6 +24,7 @@
 package com.cloudbees.workflow.rest.endpoints;
 
 import com.cloudbees.workflow.Util;
+import com.cloudbees.workflow.flownode.FlowNodeUtil;
```
kzantow: ditto

rest-api/src/main/java/com/cloudbees/workflow/util/ModelUtil.java
```diff
@@ -56,6 +56,6 @@ public static String getFullItemUrl(String itemUrl) {
 
     public static String getRootUrl() {
         StaplerRequest currentRequest = Stapler.getCurrentRequest();
-        return currentRequest.getContextPath();
+        return (currentRequest != null) ? currentRequest.getContextPath() : "/";
```
kzantow: should this be `currentRequest.getContextPath() + "/"` ?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,108 @@
...
+            if ( myJenkins == null) {
+                return FALLBACK_CACHES;
+            } else {
+                return myJenkins.getExtensionList(CacheExtension.class);
+            }
+        }
+    }
+
+    @CheckForNull
+    public static RunExt getCachedRun(FlowExecution ex) {
```
amuniz: Is this method always returning `null`?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -177,31 +279,52 @@ private static FlowNode getStageExecStartNode(List<FlowNode> allNodesSorted, int
      * @return The name of the node on which the supplied FlowNode executed.
      */
     public static String getExecNodeName(FlowNode flowNode) {
-        String execNodeName = FlowNodeExecutorNameCache.getExecNodeName(flowNode);
+        if (flowNode == null) {
```
amuniz: Under what conditions is `flowNode == null`?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -249,20 +379,31 @@ public static FlowNode getStageEndNode(FlowNode stageStartNode) {
...
     public static FlowNode getFlowEndNode(FlowExecution execution) {
+        if (execution == null) {
+            return null;
+        }
+
         // If there's no next stage, then use the last node in the workflow.
         List<FlowNode> endNodes = new ArrayList<FlowNode>(execution.getCurrentHeads());
         sortNodesById(endNodes);
 
+        // We do not actually need to sort this, we can just the get the head of the execution
```
amuniz: This comment is confusing. `endNodes` is sorted just two lines before, so I guess it needed to be sorted.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -297,90 +451,87 @@ public static FlowNode getFlowEndNode(FlowExecution execution) {
...
+        if (isCacheable && executionUrl != null) {
+            cache.put(executionUrl, nodes);
+        }
+        return nodes;
     }
 
-    static List<FlowNode> getUnsortedSortedExecutionNodeList(List<FlowNode> nodeList) {
-        if (nodeList == null || nodeList.isEmpty()) {
-            return Collections.EMPTY_LIST;
+    @Extension
```
amuniz: Clearer in line 497 IMO

rest-api/src/main/java/com/cloudbees/workflow/util/ModelUtil.java
svanoort: No, it's a placeholder here, to allow building a dummy URL string (for testing) when there is not a request context.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: If you are using the util without a running Jenkins instance, yes (example: unit tests).

Otherwise it always returns the extension.

amuniz: Then you have missed a `return` somewhere.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @amuniz Sorry, I do not understand what you mean?

amuniz: Nothing important, just that `@Extension` on top of the javadoc is confusing, this:

```java
/** 
 * This is used to cover an obscure case where a WorkflowJob is renamed BUT
 * a previous WorkflowJob existed with cached execution data.
 * Otherwise the previous job's cached data would be returned.
 */
@Extension
public static class RenameHandler extends ItemListener {
```

would be more readable.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Oh, in getCachedRun!  Oops, that was broken when I pulled caches into an Extension.  This is why late night coding is a Bad Idea.  Fixed in latest commit, thanks

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
svanoort: Fixed, thanks - left over from before I pulled out cache invalidation at Jesse's request.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @amuniz Ah, I see - yes, you are right, changed in latest commit

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Agreed, changed in latest commit.   This will all go away with #6 though, almost all the stage analysis is restructured completely and most of the methods become needless.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Just a protective nullcheck, since it's a public API.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
     private FlowNodeUtil() {
     }
 
-    public static long getNodeExecDuration(FlowNode node) {
-        List<FlowNode> childNodes = getChildNodes(node);
-        if (!childNodes.isEmpty()) {
-            long startTime = TimingAction.getStartTime(node);
-            long endTime = TimingAction.getStartTime(childNodes.get(childNodes.size() - 1));
+    public abstract static class CacheExtensionPoint implements ExtensionPoint {
```
kzantow: prefer something like `FlowNodeCache...` and similar naming for the `CacheExtension` below, even though these are nested classes...

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
...
+        public abstract Cache<String,List<FlowNode>> getExecutionCache();
+        public abstract Cache<String, RunExt> getRunCache();
+        public abstract Cache<FlowNode,String> getExecNodeNameCache();
+    }
 
-            if (startTime == 0) {
-                // The node is running and the time has not been marked on it yet.  Return 0 as the duration for now.
-                return 0;
+    // Used in testing where Jenkins is not running yet
+    private static final List<CacheExtension> FALLBACK_CACHES = Arrays.asList(new CacheExtension());
```
kzantow: Is this actually returned during testing? Wouldn't this be a similar issue to having the static caches that were here before the extension point was introduced?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
...
+                return myJenkins.getExtensionList(CacheExtension.class);
+            }
+        }
+    }
+
+    @CheckForNull
+    public static RunExt getCachedRun(FlowExecution ex) {
+        try {
+            if (ex != null) {
+                RunExt cachedRun = CacheExtension.all().get(0).getRunCache().getIfPresent(ex.getUrl());
```
kzantow: 🐜 every use of this is `CacheExtension.all().get(0)` why not just have a `CacheExtension.get()` returning the single instance?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @kzantow Yep, it's hit in the FlowNodeUtilTest - the other tests don't suffer from issues with a static cache because they have running Jenkins instances.  However, while the FlowNodeUtil test does hit the request to get cache classes, it never actually adds anything to the cache, because there is no actual FlowExecution, and it is ineligible for caching.   A bit esoteric, I know.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @kzantow Yeah, it's a little bit awkward, no lie.  Rather than addressing this in this PR though, let me take care of it with https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/6 (I've added a TODO to my task list there).  I'm stripping out most of the caches there (no longer needed when the new FlowAnalyzer is added, except for the Run cache).

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -177,31 +284,52 @@ private static FlowNode getStageExecStartNode(List<FlowNode> allNodesSorted, int
...
                     execNodeName = node;
                 }
+                break;
+            }
+
+            // Next look at ancestors
+            ancestry.push(flowNode);
+            List<FlowNode> parents = flowNode.getParents();
+            if (parents != null && !parents.isEmpty()) {
+                flowNode = parents.get(0);
```
kzantow: Does it matter if there are multiple parents? Guess not really...

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -249,12 +384,17 @@ public static FlowNode getStageEndNode(FlowNode stageStartNode) {
         return allNodesSorted.get(allNodesSorted.size() - 1);
     }
 
+    @CheckForNull
```
kzantow: 🐜  location ⬇️

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
rsandell: The list is not really a constant, since `Arrays.asList` returns an `ArrayList` that can be added and removed from, but since it's `private` I see no real harm with it, otherwise it would need to be wrapped with `Collections.unmodifiableList`

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
rsandell: Why is an `ExtensionPoint` a nested class? It smells a bit *iffy* to me.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
rsandell: Given my comment above that the `FALLBACK_CACHES` can be cleared you run a risk of getting an index out of bounce here.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
rsandell: Then it should be annotated as well.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Multiple parents only apply to parallel stages; this approach is not smart enough to handle that anyway, which is one of the things being addressed in the new FlowAnalyzer.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -276,10 +421,23 @@ public static FlowNode getFlowEndNode(FlowExecution execution) {
         return nodes;
     }
 
+    // Enables us to get the status of a node without creating a bunch of objects
+    public static StatusExt getStatus(FlowNode node) {
+        boolean isExecuted = NotExecutedNodeAction.isExecuted(node);
+
+        if (isExecuted) {
+            ErrorAction errorAction = node.getError();
+            return StatusExt.valueOf(errorAction);
```
kzantow: :ant: `@CheckForNull` / `NPE` here?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Thanks, noted, +1 on this, if this gets another commit (beyond merges from master), I will move it, otherwise, will do in the other PR.

pom.xml
```diff
@@ -33,7 +33,7 @@
 
     <groupId>org.jenkins-ci.plugins.pipeline-stage-view</groupId>
     <artifactId>parent-pom</artifactId>
-    <version>1.2-SNAPSHOT</version>
+    <version>1.2-beta-2-SNAPSHOT</version>
```
jglick: Why?

Changing versions around is fine, if you prefer to use `--batch` when running releases rather than typing in custom numbers, but best not to do as part of a feature PR.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeListCacheAction.java
jglick: Did you specifically make sure the old builds were loaded, for example by displaying the stage view for them? Builds, flow nodes, and flow node actions are all loaded lazily.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeListCacheAction.java
```diff
@@ -1,81 +0,0 @@
...
-    private static final Logger LOGGER = Logger.getLogger(FlowNodeUtil.class.getName());
-
-    public transient List<FlowNode> unsortedNodeList;
-    public transient List<FlowNode> idSortedNodeList;
-
-    public static FlowNodeListCacheAction get(FlowNode flowNode) {
-        Queue.Executable execution;
-        try {
-            execution = flowNode.getExecution().getOwner().getExecutable();
-        } catch (NullPointerException e) {
```
jglick: Happy to see this nonsense be deleted! :-)

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: Why is it `abstract`? You only need only implementation, right? So it would be simpler to just have a single `final` class.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: Well, `Collections.singletonList` is clearer and also unmodifiable.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
...
-            if (startTime == 0) {
-                // The node is running and the time has not been marked on it yet.  Return 0 as the duration for now.
-                return 0;
+    // Used in testing where Jenkins is not running yet
+    private static final List<CacheExtension> FALLBACK_CACHES = Arrays.asList(new CacheExtension());
+
+    @Extension
+    public static class CacheExtension extends CacheExtensionPoint {
+
+        protected final Cache<String,List<FlowNode>> executionCache = CacheBuilder.newBuilder().weakValues().maximumSize(100).build();
```
jglick: Why not `private`?

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
```diff
@@ -183,11 +184,49 @@ public void setArtifacts(Link artifacts) {
...
+        }
+
+        if (!runExt.getStages().isEmpty()) {
+            // We're done when the last stage is, and it is the result of the overall run
+            FlowNodeExt lastStage = runExt.getLastStage();
+            runExt.setEndTimeMillis(lastStage.getStartTimeMillis()+lastStage.getDurationMillis());
+            lastStage.setStatus(runExt.getStatus());
+        }
+
+        long currentTimeMillis = System.currentTimeMillis();
```
kzantow: Using `currentTimeMillis` here seems wrong... determining any execution timing info here seems wrong. It looks like this is just to deal with in progress executions?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: Yes, better to have a `get()` method returning the singleton.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: :ant: this would better be an instance method on `CacheExtension` itself, to prepare the way for threading the cache through other extensions, so the static getter eventually goes away.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: …especially since you are elsewhere referring to `CacheExtension`, not `CacheExtensionPoint`! :ant:

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: On the extension?  It wasn't legal if scoped to protected, I would assume even less legal if scoped to prive.

jglick: :ant: the use of `weakValues` with a `List` means that the cache is not very useful, since every GC cycle will throw away all the entries. The right approach is to use strong values but cache a `List<String>` IDs, where the actual `FlowNode`s are looked up on demand by ID; or to actually include all the metadata you need in some struct class in the value.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
...
+
+        // Larger cache of run data, for completed runs, keyed by flowexecution url, useful for serving info
+        // Actually can be used to serve Stage data too
+        // Because the RunExt caps the total elements returned, and this is fully realized, this is the fastest way
+        protected final Cache<String, RunExt> runData = CacheBuilder.newBuilder().maximumSize(1000).build();
```
jglick: :ant: better `private`

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
...
+            if (ex != null) {
+                RunExt cachedRun = CacheExtension.all().get(0).getRunCache().getIfPresent(ex.getUrl());
+                if (cachedRun != null) {
+                    return cachedRun;
+                }
+            }
+            return null;
+        } catch (IOException ioe) {
+            LOGGER.severe("Can't get execution url for execution, IOException!");
+            throw new IllegalStateException(ioe);
```
jglick: :ant: inappropriate and undeclared exception, not to mention pointless logging. The method is `@CheckForNull`, so you should

```java
LOGGER.log(Level.WARNING, null, ioe);
return null;
```

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
svanoort: @kzantow Correct, it's just for in-progress executions.  Although the diff shows this as new, it is just refactored out from the previous createMethod (to allow for the new FlowAnalyzer to invoke it separately from the other creation routines).

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
...
+        }
+    }
+
+    public static void cacheRun(FlowExecution exec, RunExt run) {
```
jglick: :ant: Again this should be an instance method.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -49,28 +71,113 @@
...
+        if (exec != null && exec.isComplete()) {
+            try {
+                CacheExtension.all().get(0).getRunCache().put(exec.getUrl(), run);
+            } catch (IOException ioe) {
+                LOGGER.severe("Can't get execution url for execution, IOException!");
```
jglick: :ant: useless log message since the stack trace is not being recorded.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
kzantow: Right, the weak reference thing seems like an issue on a heavily used Jenkins instance, can just use LRU with a reasonable maximum on this cache, no?

svanoort: @jglick Already addressed two different ways.  First, the whole RunExt gets cached once the run and the run analysis are complete; this execution cache is only used while processing the run. So, it only needs to live for a few seconds in the cache to get the full use (with 10k+ FlowNodes under analysis, about 5-6 seconds on my box).  

Second, in the parent methods that generate the run, we obtain and hold a strong ref to the List while the analysis is progressing - https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/4/commits/50d093a5f7d4aa1591572d2735635ecbb95e8a63#diff-66fea105c6bbbe32760bdb0c2a604c10R336

Worst case, even if it weren't strong-referenced, it would just mean we ran slower because we wouldn't get values from cache.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Isn't it a pretty serious issue if we can't get the URL of a correctly executed FlowExecution though? IIRC that should only happen in case where the data is corrupt on disk.

svanoort: @jglick Worth mentioning I waffled on this, a previous version logged (at SEVERE) and returned a null (maintaining previous behavior in places), but I realized that if this couldn't get the URL it was probably a symptom of a bigger major issue.   Willing to waffle it back though (probably in the successive PR).

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: Agreed, either annotate the parameter as `@CheckForNull` (and specify in Javadoc what it means for the node to be null), or mark it `@Nonnull` and do not runtime check and make sure callers comply.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -177,31 +284,52 @@ private static FlowNode getStageExecStartNode(List<FlowNode> allNodesSorted, int
...
+                flowNode = parents.get(0);
             } else {
-                execNodeName = getExecNodeName(parent);
+                break;
             }
         }
-
-        FlowNodeExecutorNameCache.cache(execNodeName, flowNode);
-
+        // Ran on master if no executor workspace
```
jglick: :ant: If there was no workspace, it did not run on master, it just did not run in a node at all. (A step run in a `node` block on the master will have a real node name, `""`.) So this looks wrong. Probably it should be returning null.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeListCacheAction.java
svanoort: @jglick I was even happier to remove it! :-)   There's a valid case for using an action to cache to disk in a *very* few cases, but this is not it.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -249,12 +384,17 @@ public static FlowNode getStageEndNode(FlowNode stageStartNode) {
...
     }
 
+    @CheckForNull
     /**
      * Get the last node to start in a flow.
      * @param execution The flow execution.
      * @return The last node to start in the flow.
      */
     public static FlowNode getFlowEndNode(FlowExecution execution) {
+        if (execution == null) {
```
jglick: :ant: again use FindBugs and decide whether you intend to allow a parameter to be null or not.

jglick: Generally methods like this should take a `@Nonnull` parameter.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -263,6 +403,11 @@ public static FlowNode getFlowEndNode(FlowExecution execution) {
     }
 
     public static List<FlowNode> getStageNodes(FlowExecution execution) {
+
+        if (execution == null) {
```
jglick: :ant: ditto

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -297,90 +455,88 @@ public static FlowNode getFlowEndNode(FlowExecution execution) {
...
-        FlowGraphWalker walker = new FlowGraphWalker(parentNode.getExecution());
-        for (FlowNode node : walker) {
-            if (node.getParents().contains(parentNode) && !nodes.contains(node)) {
-                nodes.add(node);
+        String executionUrl = null;
+        boolean isCacheable = false;
+        try {
+            executionUrl = execution.getUrl();
+        } catch (IOException ioe) {
+            LOGGER.severe("Can't get execution url for execution, IOException!");
```
jglick: ditto

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeListCacheAction.java
svanoort: @jglick yes, I visited them in stage view.

kzantow:
I'm a little dubious about the "reference" held on to the flow node list vs. the `executionCache` weak reference, especially when the vm optimizer kicks in... (e.g. that variable is never read, might be eligible to optimize out, but I really have no idea of the behavior; unsure if that would end up causing a big performance issue or not, anyway).

I don't see any blockers here, though, so :bee:

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
```diff
@@ -332,8 +332,8 @@ public static RunExt createOld(WorkflowRun run) {
             if (artifacts != null && !artifacts.isEmpty()) {
                 runExt.get_links().setArtifacts(Link.newLink(RunAPI.getArtifactsUrl(run)));
             }
-
             FlowGraphWalker walker = new FlowGraphWalker(execution);
+            List<FlowNode> sortedNodes = FlowNodeUtil.getIdSortedExecutionNodeList(execution); // Hold a ref to prevent GC until we're done analyzing
```
kzantow: I'm not convinced this would actually prevent it from being removed from the cache.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Please look closely at the original code here - this is the expected handling, you might not agree with it (I can see the logic), but it is not wrong, within the context of this PR (changing that behavior would be out of scope). 

That said, the original implementation (which this is only improving scaling on, until the new FlowAnalyzer is read), was patently broken and did not even attempt to look at block scoping, so... the entirety here is going to be ripped out and we can discuss if the behavior should change then.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @kzantow Jesse has stated that we are not allowed hold Strong references to an Item - I am still waiting for the explanation behind it, but see comment above for how I am avoiding GC (or trying to).

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -297,90 +455,88 @@ public static FlowNode getFlowEndNode(FlowExecution execution) {
...
-                int node2Iota = parseIota(node2);
-
-                if (node1Iota < node2Iota) {
-                    return -1;
-                } else if (node1Iota > node2Iota) {
-                    return 1;
-                }
+        private int parseIota(FlowNode node) {
+            try {
+                return Integer.parseInt(node.getId());
```
jglick: This is a bug, but predates this PR.

rest-api/src/main/java/com/cloudbees/workflow/rest/AbstractAPIActionHandler.java
```diff
@@ -51,5 +51,4 @@ public String getDisplayName() {
         // No display
         return null;
     }
-
```
jglick: ?

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/FlowNodeAPI.java
```diff
@@ -31,9 +31,6 @@
 import hudson.Extension;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 
-import java.util.ArrayList;
-import java.util.List;
-
```
jglick: ?

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -62,8 +62,8 @@ public void test() throws Exception {
                 "   stage ('Test'); " +
                 "   echo ('Testing'); " +
                 "   stage ('Deploy'); " +
-                ( hudson.Functions.isWindows() ? "   bat ('rmdir /s/q targs || echo no such dir\\n mkdir targs && echo hello> targs\\\\hello.txt'); "
-                                               : "   sh ('rm -rf targs && mkdir targs && echo hello > targs/hello.txt'); "
+                (hudson.Functions.isWindows() ? "   bat ('rmdir /s/q targs || echo no such dir\\n mkdir targs && echo hello> targs\\\\hello.txt'); "
+                        : "   sh ('rm -rf targs && mkdir targs && echo hello > targs/hello.txt'); "
```
jglick: BTW you can use steps like `deleteDir` & `writeFile` to avoid a platform dependency.

jglick:
No clue what most of the code is doing, and left various notes, but :bee: overall.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: There _might_ be a serious issue, but the best way to recover gracefully is to the log the exception, and continue with some sensible default behavior so at least you are not throwing `IllegalStateException`s in the user’s face.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: > Jesse has stated that we are not allowed hold Strong references to an Item - I am still waiting for the explanation behind it

Bad, bad things happen.

jglick: > On the extension?

No, on the field. Fields should be `private` unless there is a good reason to make them otherwise. And `protected` should only ever be used on classes you intend to subclass.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
svanoort: It depends on how smart the JIT compiler is feeling here.  I also am not 100% sure it won't strip it unless I poke the list a bit more; but in any case, we've got a fallback and a fallback to the fallback, and remember this entire cache is being stripped out soon.

kzantow: Right. These days they are pretty smart. And I've seen things optimized out that were even referenced in parts of unused branch expressions. Anyway, as you say, it seems the fallback(s) will at least let this continue to function in this case. I was just noting it to be reviewed subsequently.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Sure does!  You'll be happy to know this is being addressed in the new FlowAnalyzer.

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/FlowNodeAPI.java
svanoort: Cleanup, since I was cleaning up other depts.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
svanoort: Noted - this one was inherited, to be honest I'm not sure why there's even a whitespace change here since I don't remember touching this, but... after this many commits, who knows.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: This whole implementation is iffy to me.  We're creating an ExtensionPoint & Extension solely to hack around the fact that the JenkinsRule does classloading wrong (the same classloader is used across tests).  The initial implementation with the static cache + invalidation was simpler and more straightforward.  But I raised my protest about that already and was overruled, so if it's going to be ugly, at least it's be confined to a few parts of one file. 

I think that there are far better ways to do this overall, but rather than try to put a dress and bowties on a pig, I'm going to save it for architectural changes along with the new FlowAnalyzer (hopefully I can find a place where it's logical and easy to put the cache, rather than in the Util class).

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @jglick I'd really like to know what flavor of bad, in case it rears its head in the future... :wink: 

I am not ready to close this for extension yet, ergo the protected, but if the next iteration of algorithms doesn't touch it, it'll get locked down probably.

rest-api/src/main/java/com/cloudbees/workflow/rest/AbstractAPIActionHandler.java
svanoort: I feel the same way.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
jglick: Well as of JDK 8u74 if you comment out [this line](https://github.com/jenkinsci/jenkins/blob/511ccf0c836651d1d825c062d7d42338438c735c/test/src/test/java/hudson/diagnosis/OldDataMonitorTest.java#L101) the test fails, which I take as evidence that at least the first-pass compiler is _not_ that aggressive. Whether the optimized version produced after you run the method a few thousand times tries to discard the reference, I do not know.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: It _is_ simple and straightforward if you do it right. You just need one extension, with all the state.

```java
@Extension public final class FlowNodeCache {
    public static FlowNodeCache get() {
        return ExtensionList.lookup(FlowNodeCache.class).iterator().next();
    }
    private final Cache<String,Integer> exampleField = …;
    public int count(String key) {return exampleField.get(key);}
}
```

In the short term, callers can use `get()`:

```java
int count = FlowNodeCache.get().count("hello");
```

But if the caller is itself an extension, it is easier:

```java
@Extension public class Whatever extends RootAction {
    @Inject private FlowNodeCache cache;
    public void doStuff(@QueryParameter String stuff) {
        int count = cache.count(stuff);
        // …
    }
}
```

and when the caller is not an extension, you just thread the `FlowNodeCache` singleton instance through your method calls.

Unit tests need merely construct an empty cache object and pass them around.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
```diff
@@ -209,8 +248,92 @@ public static RunExt create(WorkflowRun run) {
...
+            List<StageNodeExt> myWrappedStages = new ArrayList<StageNodeExt>();
+            if (wrappedStages == null) {
+                for(StageNodeExt stage : run.getStages()) {
+                    myWrappedStages.add(stage.myWrapper());
+                }
+                this.wrappedStages = myWrappedStages;
+            } else {
+                wrappedStages = null;
+            }
+        }
```
jglick, line 288: AFAICT this does not work: `wrappedStages` is always initialized as a copy of `run.stages`, meaning nothing is in fact hidden.

PR #5, opened 2016-04-01
[JENKINS-33110] Log details not shown

amuniz:
[JENKINS-33110](https://issues.jenkins-ci.org/browse/JENKINS-33110)

- [x] Expand log sections on click (regression)
- [x] When there are more than 15-20 steps in the same stage the log is not shown for last steps in the list (the calculation of the height was overlapping previous layers in the DOM)
- [x] Adjust the log DOM container to its parent element and make the lines match with the background
- [x] Test

@reviewbybees esp. @tfennelly @recena


recena:
:bee: since I participated in the PR

svanoort:
:bee:

PR #6, opened 2016-04-04
[WIP] New flow analyzer

svanoort:
**DO NOT REVIEW, THIS IS CHANGING RAPIDLY**

This bases on top of https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/4 but pulls out components needing additional work to prepare.  

# What this offers
- [x] Greatly increased flow scanning performance
- [ ] Refactor needed: split a flow into FlowSegments what are linear paths (which can have forks within them), which handles parallel branches.  Some refactors to store this more generally/gracefully vs. current Stage mechanism
- [ ] Track parallel branches correctly [JENKINS-33185](https://issues.jenkins-ci.org/browse/JENKINS-33185) - parallel FlowSegments here.
- [ ] Correctly assign the executor used to the nodes and stages by tracking blocks and block scoping. Fixes [JENKINS-33290](https://issues.jenkins-ci.org/browse/JENKINS-33290)  -- *We need to push/pop to a Stack (actually an ArrayDeque) of block scopes, upon encountering nodes that are instanceof BlockStartNode and BlockStopNode*
- [ ] Filter options for whether or not to return child flow nodes within a linear view

# Architecture notes
- This tracks FlowSegments, which are defined as the linear flow execution portions.  All FlowNodes are mapped to one FlowSegment, and FlowSegments may be nested.  FlowSegments have first & last node, preceding and next segments, and durations.  They may also be nested (to represent blocks nested in other blocks, or stages)
- This is a SAX-like handler, it calls implementation methods to hand the start/stop of a linear flow segment, start/stop a set of parallel forks, filter/handle a node to track or note within a FlowSegment, start/stop a block
- This is based on iterating *once* through a flow graph, rather than collecting all points, sorting, and searching back and forth to compute statistics.

# Restoring stage
- [ ] Restore content removed from PR #4 in 91bceea08417f983980c0a28170d5854ba848b56
  - [ ] New FlowAnalyzer and uses of it
  - [ ] CacheStatsExt & API to manage it and invalidate the cache
  - [ ] Benchmark & HugeFlow test
  - [ ] TestFlowAnalyzer test 

# TODO
- [ ] Assign nodes to single flow segments, and track branches independently rather than merging into parent flow -> requires separating stages into sets of FlowSegment entries
- [ ] Every use of this is CacheExtension.all().get(0) why not just have a CacheExtension.get() returning the single instance.  
- [ ] Nest FlowSegment branches within FlowSegment
- [ ] Replace the nodecount check on a FlowSegment with a filter method (FlowSegment, FlowNode)
- [ ] APIs that trigger on entering/leaving a block

# Tests
- [ ] Duration/timing testing in depth, and with parallel branches
- [ ] Assignment of executor node to stage nodes not inside a node block
- [ ] Nesting of FlowSegments

# Extensibility
- [ ] Accept predicates to filter nodes in a FlowSegment


svanoort:
This is not actually merged.  Go home Github, you're drunk.

PR #7, opened 2016-04-06
Dependency fixes

jglick:
Fixes a linkage error when updating `workflow-support` to 2.0.

Critical, but this change is disruptive enough that we may want to hold off on it. I will in parallel work on an alternate compatibility fix.

@reviewbybees esp. @svanoort


pom.xml
```diff
@@ -46,8 +46,7 @@
     </scm>
 
     <properties>
-        <jenkins.version>1.609.3</jenkins.version>
-        <workflow.version>1.14</workflow.version>
+        <jenkins.version>1.642.3</jenkins.version>
```
svanoort: Why are we bumping to 1.642.3 jenkins dependency?  This removes usability for many customers.

ui/gulpfile.js
```diff
@@ -46,3 +46,4 @@ builder.bundle('src/main/js/stageview.js')
 // Explicitly setting the task list so as to disable jshint
 // TODO: Remove the line below + fix jshint build errors
 builder.defineTasks(['test', 'bundle', 'rebundle']);
+builder.defineTask('lint', function() {}); // https://github.com/jenkinsci/plugin-pom/pull/20#issuecomment-206374472
```
svanoort: Awesome.  I was building a PR that included this.

pom.xml
amuniz: What was producing the linkage error? And why it needs to bump baseline to 1.642.3?

pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.3</version>
+        <version>2.6</version>
```
svanoort: This I like.  I have a commit adding it in a branch I wanted to push.

pom.xml
amuniz: Ok, I think the answer to my second question is because Pipeline 2.0 requires that baseline.

jglick: Correct.

svanoort:
@jglick Can this be closed now that the other PR is merged?

jglick:
No, this is still useful for the future, because the `optional` dependency is just a hack to allow the class loader linkage to be made. It does not, for example, block you from disabling `pipeline-input-step`, and the current `rest-api` code is not prepared for the possibility that this dependency is missing. So either we keep the policy of a hard dep on the presence of the `input` step, in which case something like this PR is needed as soon as we are comfortable increasing the core dependency; or the processing of `input` is made into a proper extension, either loaded optionally from `pipeline-rest-api`, or factored into a separate plugin in the style of `cloudbees-workflow-ui`.

jglick:
> block you from disabling `pipeline-input-step`

Checked this in 1.642.3 and the UI _does_ block you from disabling it, despite the optional status of the dependency. Perhaps a core bug fixed later?

svanoort:
@jglick The next stage view release will retain the 1.609.x dependency, then the one after it will take up a version of this along with dependency on 1.642.x in order to use the new FlowScanner algorthms.

svanoort:
This is rolled up into https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/25 now.

PR #9, opened 2016-04-07
[WIP] New flow walker mk2 [JENKINS-34038]

svanoort:
Since GitHub "helpfully" auto-closed #6 thinking it was merged...

**DO NOT REVIEW, THIS IS CHANGING RAPIDLY**

This bases on top of https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/4 but pulls out components needing additional work to prepare.  

Implements [JENKINS-34038](https://issues.jenkins-ci.org/browse/JENKINS-34038)

Architectural notes (this keeps switching around):
* Best way to structure is to provide an object that collects stats as branches, nodes, stages, etc are visited
* Breadth-first search is actually simpler to implement (and requires less RAM due to parent refs) than depth-first search where we care about branches. 
* Provide a method that iterates through a linear (single-parent) line of nodes efficiently.  We only need to know the parent node. 
* Provide method/while loops that:  do nextNode, nextBlockStart, nextEnclosingLabel (block or stage) / nextEnclosing (block start or stage marker, jumping over enclosed blocks). 
* Need to collect stats in an internal object with state modified by iteration
* For performance, it is helpful to *not* keep a hashset of all visited nodes. O(1) lookup still requires hashing each.  Better to use breadth-first search, and stop when parallel branches end. 
     - In this implementation, when we encounter a branch we would: 
         1. Get the node beginning the (actually the end node for search) parallel block, and set it as endNode with. 
         2. Create an ArrayDeque of branches to search
         3. For each node, do a linear search within the current block until we hit the case where branches merge, using the == (identity) comparator

# What this offers
- [x] Greatly increased flow scanning performance
- [ ] Refactor needed: split a flow into FlowSegments what are linear paths (which can have forks within them), which handles parallel branches.  Some refactors to store this more generally/gracefully vs. current Stage mechanism
- [ ] Track parallel branches correctly [JENKINS-33185](https://issues.jenkins-ci.org/browse/JENKINS-33185) - parallel FlowSegments here.
- [ ] Correctly assign the executor used to the nodes and stages by tracking blocks and block scoping. Fixes [JENKINS-33290](https://issues.jenkins-ci.org/browse/JENKINS-33290)  -- *We need to push/pop to a Stack (actually an ArrayDeque) of block scopes, upon encountering nodes that are instanceof BlockStartNode and BlockStopNode*
- [ ] Filter options for whether or not to return child flow nodes within a linear view

# Architecture notes
- This tracks FlowSegments, which are defined as the linear flow execution portions.  All FlowNodes are mapped to one FlowSegment, and FlowSegments may be nested.  FlowSegments have first & last node, preceding and next segments, and durations.  They may also be nested (to represent blocks nested in other blocks, or stages)
- This is a SAX-like handler, it calls implementation methods to hand the start/stop of a linear flow segment, start/stop a set of parallel forks, filter/handle a node to track or note within a FlowSegment, start/stop a block
- This is based on iterating *once* through a flow graph, rather than collecting all points, sorting, and searching back and forth to compute statistics.

# CHANGES FROM JESSE DISCUSSION / FLOWSCANNER WORK
- [ ] Don't do flowsegment, do Block 
- [ ] For blocks, track if it contains any parallel?
- [ ] APIs for block handling:

```java
public interface BlockStorage {
  public Block enclosingBlock(FlowNode node);
 public Block[ ] scopes(FlowNode node);

  /** Nearest block where startNode matches predicate */
  public Block nearestMatchingBlock(FlowNode node, Predicate<FlowNode>); 
}

public static class Block {
  String start;  // Start/Stop block nodes
  String end;
  String firstChild;  // Enclosed nodes
  String lastChild;
  boolean hasBranches; //Does it contain parallel
  boolean contains(FlowNode node); //Can use a range query if there are no parallel sibling blocks to this block, I.E. is the iota between firstChild and lastChild.
  boolean containsId(int nodeId); 
  // TBD: store previous/next in linkedlist style?
 // TBD: store parent/child
  // Maybe these
  long startTime;
  long endTime;
  StatusExt status; 
  int childCount;
}
```

# Restoring stage
- [x] Restore content removed from PR #4 in 91bceea08417f983980c0a28170d5854ba848b56
  - [x] New FlowAnalyzer and uses of it
  - [ x CacheStatsExt & API to manage it and invalidate the cache
  - [x] Benchmark & HugeFlow test
  - [x] TestFlowAnalyzer test 
  - [x] Vet content and verify that we didn't restore anything we oughtn't to have
  - [x] Re-do any cleanups done in in the previous pullout of new flow analyzer for #4 
  - [ ] Apply the appropriate changes from #4 :ant:  comments
# TODO
- [ ] Assign nodes to single flow segments, and track branches independently rather than merging into parent flow -> requires separating stages into sets of FlowSegment entries
- [x] Every use of this is CacheExtension.all().get(0) why not just have a CacheExtension.get() returning the single instance.  
- [ ] Nest FlowSegment branches within FlowSegment
- [ ] Replace the nodecount check on a FlowSegment with a filter method (FlowSegment, FlowNode)
- [ ] APIs that trigger on entering/leaving a block

# Tests
- [ ] Duration/timing testing in depth, and with parallel branches
- [ ] Assignment of executor node to stage nodes not inside a node block
- [ ] Nesting of FlowSegments

# Extensibility
- [ ] Accept predicates to filter nodes in a FlowSegment


svanoort:
Closing this because https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/25 is the spiritual descendant and does this in a much more sane way.

PR #14, opened 2016-06-08
Properties to configure hardcoded limits on APIs - JENKINS-34791

svanoort:
Provides java system properties to configure the hardcoded limits on jobs per run, etc per [JENKINS-34791](https://issues.jenkins-ci.org/browse/JENKINS-34791).

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/JobExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/JobExt.java
@@ -42,7 +42,7 @@ public class JobExt {
     /**
      * Max number of runs per page. Pagination not yet supported.
      */
-    public static final int RUN_PAGE_SIZE = 20;
+    public static final int MAX_RUNS_PER_JOB = Integer.getInteger(JobExt.class.getName()+".maxRunsPerJob", 10);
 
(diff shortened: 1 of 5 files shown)


svanoort:
@reviewbybees

jglick:
I recommend filing PRs from forks rather than origin branches.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/JobExt.java
```diff
@@ -42,7 +42,7 @@
     /**
      * Max number of runs per page. Pagination not yet supported.
      */
-    public static final int RUN_PAGE_SIZE = 20;
+    public static final int MAX_RUNS_PER_JOB = Integer.getInteger(JobExt.class.getName()+".maxRunsPerJob", 10);
```
jglick: Why the reduction?

jglick: I see, it was already capped in `RunGroupGenerator`.

jglick:
:bee:

oleg-nenashev:
:bee:

cgomestw:
have a way to get more runs via curl without change env variable?

PR #11, opened 2016-04-29
Fix flow node links when used with no Stapler request

jimmidyson:
Integrating with other plugins that need to get a JSON representation of the
pipeline state there is no Stapler request. Currently this throws an NPE which
this fixes.

PR #15, opened 2016-06-09
Property to disable user lookup for changesets for JENKINS-35484

svanoort:
Resolves [JENKINS-35484](https://issues.jenkins-ci.org/browse/JENKINS-35484). 

Verified manually by changing the property using groovy calls.

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
@@ -27,6 +27,7 @@ import com.cloudbees.workflow.util.ModelUtil;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import hudson.scm.ChangeLogSet;
 import hudson.scm.RepositoryBrowser;
+import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 
 import java.io.IOException;
@@ -51,8 +52,7 @@ public class ChangeSetExt {
      * This is a workaround for JENKINS-35484 where user lookup encounters issues */
     private static boolean resolveCommitAuthors() {
         String prop = System.getProperty(ChangeSetExt.class.getName()+".resolveCommitAuthors");
-        // Anything but something *explicitly* matching "false" (case insensitive) needs to return true
-        return (prop == null || prop.isEmpty() || !"false".equalsIgnoreCase(prop));
+        return (StringUtils.isEmpty(prop)|| Boolean.parseBoolean(prop));
     }
 
     public String getKind() {
@@ -183,7 +183,7 @@ public class ChangeSetExt {
             commit.setCommitId(entry.getCommitId());
             commit.setCommitUrl(repoUrl);
             commit.setMessage(entry.getMsg());
-            commit.setAuthorJenkinsId(resolveCommitAuthors() ? entry.getAuthor().getFullName() : "Unidentified");
+            commit.setAuthorJenkinsId(resolveCommitAuthors() ? entry.getAuthor().getFullName() : "");
             commit.setTimestamp(entry.getTimestamp());
 
             if (commit.getTimestamp() > -1) {
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
@@ -45,6 +45,9 @@ public class ChangeSetExt {
     private List<Commit> commits;
     private String consoleUrl; // Not a rest endpoint so not including in _links
 
+    /** False by default, workaround for JENKINS-35484 where user lookup breaks things */
+    private static boolean AVOID_RESOLVING_COMMIT_AUTHORS = Boolean.getBoolean(ChangeSetExt.class.getName()+".avoidResolvingCommitAuthors");
+
     public String getKind() {
         return kind;
     }
@@ -173,7 +176,7 @@ public class ChangeSetExt {
             commit.setCommitId(entry.getCommitId());
             commit.setCommitUrl(repoUrl);
             commit.setMessage(entry.getMsg());
-            commit.setAuthorJenkinsId(entry.getAuthor().getFullName());
+            commit.setAuthorJenkinsId((AVOID_RESOLVING_COMMIT_AUTHORS) ? "Unidentified" : entry.getAuthor().getFullName());
             commit.setTimestamp(entry.getTimestamp());
 
             if (commit.getTimestamp() > -1) {
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
@@ -47,10 +47,12 @@ public class ChangeSetExt {
 
 
     /** Allows user to disable Jenkins user lookup for commit authors
-     * By setting System Property com.cloudbees.workflow.rest.external.ChangeSetExt.avoidResolvingCommitAuthors to 'true'
+     * By setting System Property com.cloudbees.workflow.rest.external.ChangeSetExt.resolveCommitAuthors to 'false'
      * This is a workaround for JENKINS-35484 where user lookup encounters issues */
-    public static boolean isResolvingCommitAuthors() {
-        return !Boolean.getBoolean(ChangeSetExt.class.getName()+".avoidResolvingCommitAuthors");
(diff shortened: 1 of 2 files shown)


svanoort:
@reviewbybees

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
```diff
@@ -45,6 +45,14 @@
     private List<Commit> commits;
     private String consoleUrl; // Not a rest endpoint so not including in _links
 
+
+    /** Allows user to disable Jenkins user lookup for commit authors
+     * By setting System Property com.cloudbees.workflow.rest.external.ChangeSetExt.avoidResolvingCommitAuthors to 'true'
+     * This is a workaround for JENKINS-35484 where user lookup encounters issues */
+    public static boolean isResolvingCommitAuthors() {
```
amuniz: If it is a workaround (hopefully fixed in the right way later), better to keep this private so it can be removed at that point.

jglick: Agreed.

jglick: Note that Groovy `/script` can set `private` variables.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
```diff
@@ -45,6 +45,14 @@
...
+        return !Boolean.getBoolean(ChangeSetExt.class.getName()+".avoidResolvingCommitAuthors");
```
jglick: Clearer to call it `resolveCommitAuthors`  and default to true. (prefer positive sense to negative sense)

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
```diff
@@ -173,7 +181,7 @@ protected void mapFields(ChangeLogSet<? extends ChangeLogSet.Entry> changeset, W
             commit.setCommitId(entry.getCommitId());
             commit.setCommitUrl(repoUrl);
             commit.setMessage(entry.getMsg());
-            commit.setAuthorJenkinsId(entry.getAuthor().getFullName());
+            commit.setAuthorJenkinsId(isResolvingCommitAuthors() ? entry.getAuthor().getFullName() : "Unidentified");
```
jglick: Better to leave it null, and adjust UI to simply omit this field.

svanoort: @jglick I'm worried about direct API consumers that might be broken by nulling it.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
svanoort: Good point, I'll make the change.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
svanoort: Good point about making the call private -- I'll do that.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
jglick: Are we advertising this API for anyone else? I do not think so. It is intended to serve the stage view only IIUC.

svanoort: Red Hat's Openshift Team is directly consuming some of our APIs, see https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/11#issuecomment-217139203

jglick: Then use `""`, which should not break any code, and have the UI skip display of empty strings.

svanoort:
@jglick  and @amuniz  I have addressed your comments, please could you reexamine this?

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
```diff
@@ -45,6 +45,16 @@
...
     private String consoleUrl; // Not a rest endpoint so not including in _links
 
+
+    /** Allows user to disable Jenkins user lookup for commit authors
+     * By setting System Property com.cloudbees.workflow.rest.external.ChangeSetExt.resolveCommitAuthors to 'false'
+     * This is a workaround for JENKINS-35484 where user lookup encounters issues */
+    private static boolean resolveCommitAuthors() {
+        String prop = System.getProperty(ChangeSetExt.class.getName()+".resolveCommitAuthors");
+        // Anything but something *explicitly* matching "false" (case insensitive) needs to return true
+        return (prop == null || prop.isEmpty() || !"false".equalsIgnoreCase(prop));
```
jglick: :ant: More simply:

```java
return Boolean.parseBoolean(System.getProperty(ChangeSetExt.class.getName() + ".resolveCommitAuthors", "true"));
```

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
```diff
@@ -173,7 +183,7 @@ protected void mapFields(ChangeLogSet<? extends ChangeLogSet.Entry> changeset, W
             commit.setCommitId(entry.getCommitId());
             commit.setCommitUrl(repoUrl);
             commit.setMessage(entry.getMsg());
-            commit.setAuthorJenkinsId(entry.getAuthor().getFullName());
+            commit.setAuthorJenkinsId(resolveCommitAuthors() ? entry.getAuthor().getFullName() : "Unidentified");
```
jglick: :ant: recommend returning, say, `""`, and having the UI suppress display of a blank name.

jglick:
:bee: with qualifications

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChangeSetExt.java
svanoort: @jglick Intentional.  The reason for the comment here: https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/15#discussion-diff-66810938R54 is that Boolean.parseBoolean will return false for anything that doesn't match 'true' (case insensitive).  

We want the opposite, for safety: resolve commit authors unless they explicitly set the system property to 'false' (case insensitive).  That means we don't disable user lookup if they've mangled the setting (avoids some possible issues).

jglick: :-1:, follow the usage pattern every other call to boolean system properties uses: `true` is the only way to turn it on, anything else is considered false. If a user sets some random string, that is their problem.

svanoort:
@jglick Please can you give this one another look?  I've addressed your qualifications, and set up the UI to hide commit authors when not resolved

oleg-nenashev:
LGTM 🐝

ui/src/main/js/view/templates/run-changeset.hbs
```diff
@@ -1,9 +1,13 @@
 <div class="alert alert-info run-changeset-details {{kind}}-repo">
+    {{#if contributors.0.authorJenkinsId}}
     <div class="header">{{commitCount}} commit{{#ifCond commitCount '>' 1}}s{{/ifCond}} by <span class="contributor">{{contributors.0.authorJenkinsId}}</span>{{#ifCond contributorCount '>' 1}} and others{{/ifCond}}</div>
```
jglick: :ant: Would seem easier and more accurate to do

```hbs
<div class="header">
  {{commitCount}} commit{{#ifCond commitCount '>' 1}}s{{/ifCond}}
  {{#ifCond contributors.0.authorJenkinsId}}
    by <span class="contributor">{{contributors.0.authorJenkinsId}}</span>
    {{#ifCond contributorCount '>' 1}} and others{{/ifCond}}
  {{/ifCond}}
</div>
```

(possibly with some XML whitespace manipulation there)

jglick:
:bee:

ui/src/main/js/view/templates/run-changeset.hbs
svanoort: You're right, it's a bit simpler. In this case, don't think it's quite a big enough change to run through another review cycle on this, but I reserve the right to go back and make the change later.

Worth noting: that 3rd line would be better written as #if not #ifCond for handlebars.

PR #17, opened 2016-06-13
colWidths did not seem to exist, leading to malformed HTML/CSS

jglick:
Perhaps @tfennelly knows what this originally was supposed to be, but anyway I am sick of NetBeans showing this plugin as erroneous because of the bad CSS in the test HTML files!

@reviewbybees


amuniz:
`colWidths` seems to be [not used](https://github.com/jenkinsci/pipeline-stage-view-plugin/search?q=colWidths) anywhere in the plugin, so I guess it is safe to remove it.

:bee:

svanoort:
🐝  let's merge :)

PR #18, opened 2016-06-20
Remove node labels until they can correctly evaluate blocks with new FlowScanner

svanoort:
Node label uses can't be correctly computed without a block-aware and parallel-aware analysis method. 

We are removing them temporarily per [JENKINS-33290](https://issues.jenkins-ci.org/browse/JENKINS-33290) until a flow analyzer based on the ForkScanner in https://github.com/jenkinsci/workflow-api-plugin/pull/2 and the block-aware analysis APIs in https://github.com/jenkinsci/workflow-api-plugin/pull/6
can be added to replace FlowNodeUtil.


oleg-nenashev:
🐝 when it has a TODO comment in the code

oleg-nenashev:
🐝

PR #19, opened 2016-07-10
Fix JENKINS-34644 by using timing info for run to work around rare issues sues with timing and FlowEndNode

svanoort:
Workaround that solves [JENKINS-34644](https://issues.jenkins-ci.org/browse/JENKINS-34644) and tests the workaround.  

@reviewbybees


jglick:
Seems to be a reasonable workaround, but where is the fix for the actual bug?

svanoort:
@jglick The bug in pipeline support plugin is still under investigation -- I've got a lot of information gathered and am closing in on it, but it's tricky and does not happen often.  Once I've got the 

This at least removes the most obvious symptom (and in the process hardens stage view against another class of bug).

abayer:
:bee:

PR #21, opened 2016-07-12
Improve stageview status handling

svanoort:
Addresses [JENKINS-33700](https://issues.jenkins-ci.org/browse/JENKINS-33700) by adding UNSTABLE status, and also solves [JENKINS-36523](https://issues.jenkins-ci.org/browse/JENKINS-36523) and [JENKINS-34212](https://issues.jenkins-ci.org/browse/JENKINS-34212).

That said, annotated stages are probably becoming a must-have. 

- [x] Basic implementation of UNSTABLE status and correct stage statuses based on overall run status
- [ ] Unit tests for a variety of runs with different stage combinations + UNSTABLE (and using currentBuild.result

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
@@ -357,6 +357,19 @@ public class RunExt {
 
                 FlowNodeExt lastStage = runExt.getLastStage();
 
+                // Correct for some edge cases due to simplified handling of stages (just looking for executed and presence of error)
+                for (StageNodeExt stageNodeExt : runExt.getStages()) {
+                    // If stage has an error, but run succeeded... error was caught
+                    if (stageNodeExt.getStatus() == StatusExt.FAILED && (runExt.getStatus() != StatusExt.FAILED)) {
+                        stageNodeExt.setStatus(StatusExt.SUCCESS);
+                    }
+
+                    // If run is unstable, one of the stages must have set the status -- no idea which though
+                    if (runExt.getStatus() == StatusExt.UNSTABLE && stageNodeExt.getStatus() != StatusExt.NOT_EXECUTED) {
+                        stageNodeExt.setStatus(StatusExt.UNSTABLE);
+                    }
+                }
+
                 // Correct for rare cases with 0 for a FlowEndNode
                 if (lastStage.getPauseDurationMillis() < 0) {
                     lastStage.setPauseDurationMillis(0);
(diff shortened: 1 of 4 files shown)


svanoort:
@reviewbybees

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/FlowNodeAPITest.java
```diff
@@ -94,6 +95,160 @@ public void test_success_flow() throws Exception {
...
+                "stage 'first'\n" +
+                "echo 'works'\n" +
+                "stage 'second'\n" +
+                "currentBuild.result = 'UNSTABLE'\n" +
+                "echo 'ran things'\n" +
+                "stage 'end'\n" +
+                "echo 'done'"
+                ));
+
+        failUnstable.setDefinition(new CpsFlowDefinition("" +
```
rsandell: Two separate scenarios should probably be in different test cases, so you can see from the report which is failing if any.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/FlowNodeAPITest.java
```diff
@@ -94,6 +95,160 @@ public void test_success_flow() throws Exception {
...
+        oneStageCatch.setDefinition(new CpsFlowDefinition("" + // Tyrion in King's Landing
+                "stage 'Game of Thrones'\n" +
+                "try {\n" +
+                "    error('Trusted Grand Maester Pycelle')\n" +
+                "} catch (Exception e) {\n" +
+                "    echo 'It was a trap - Tyriowned!'\n" +
+                "}"
+        ));
+
+        multiStageCatch.setDefinition(new CpsFlowDefinition("" + //The One True King, Stannis
```
rsandell: dito

rsandell:
:bee:

rsandell:
:bee:

svanoort:
Limitation on this: for in-progress builds, where the currentBuild.result functionality is used, while the initial status request *may* work correctly, requests to the describe API for individual stage nodes will not (boo).  I'm not sure that could be handled without changing the request model as per [JENKINS-34036](https://issues.jenkins-ci.org/browse/JENKINS-34036) so that all initial stage info comes on the request (plus markers for the extension points, etc).

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
```diff
@@ -407,7 +421,20 @@ private void initStatus(WorkflowRun run) {
         } else if (execution.getCauseOfFailure() != null) {
             setStatus(StatusExt.valueOf(execution.getCauseOfFailure()));
         } else if (execution.isComplete()) {
-            setStatus(StatusExt.SUCCESS);
+            Result r = run.getResult();
+            if (r == Result.NOT_BUILT) {
```
jtnord: surely this should be using `Result.isWorseThan()` - at least for a certain class of Results?

svanoort: @jtnord What are you proposing?  In this case I'm doing a 1:1 mapping here (I'd do a mapping with switch/case if the types allowed for it).

stephenc:
:bee: AIUI

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
jtnord: I'm not proposing anything - merely stating that `Result` is not final - and normally comparing them would be a bug.
for instance if a plkugin introduced a different Result - lets call it "Whacky" and a pipeline set the result to "Whacky" this would set the internal ext status (whatever it is) to FAILED - when its somewhere between success and unstable.

jtnord:
:bee:

svanoort:
@reviewbybees done

PR #22, opened 2016-07-13
Fix handling of multiline strings for input box prompts [JENKINS-36661]

svanoort:
Easy fix that resolves [JENKINS-36661](https://issues.jenkins-ci.org/browse/JENKINS-36661).  We do escaping on the string and replace all end of line combinations with a \<br\> tag. 

@reviewbybees

### DIFF
--- a/ui/src/main/js/view/templates/index.js
+++ b/ui/src/main/js/view/templates/index.js
@@ -63,6 +63,12 @@ function registerHBSHelper(name, helper) {
     handlebars.registerHelper(name, helper);
 }
 
+registerHBSHelper('breaklines', function(text) {
+    text = handlebars.escapeExpression(text);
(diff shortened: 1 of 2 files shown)


rsandell:
:bee:

ui/src/main/js/view/templates/index.js
```diff
@@ -63,6 +63,12 @@ function registerHBSHelper(name, helper) {
     handlebars.registerHelper(name, helper);
 }
 
+registerHBSHelper('breaklines', function(text) {
+    text = handlebars.escapeExpression(text);
+    text = text.replace(/(\r\n|\n|\r)/gm, '<br>');
```
jglick: BTW there is a utility in `Functions` (or is it `Util`?) which does just this, except better (considering good break points etc.); but it is not available for JavaScript code.

svanoort:
@reviewbybees done

PR #24, opened 2016-07-30
Redo cache invalidation (bugfix, enhancement, and precursor to new flow scanning)

svanoort:
* Initially intended to address [JENKINS-36865](https://issues.jenkins-ci.org/browse/JENKINS-36865) by listening for deleted jobs and removing them from the cache.  
* Along the way, also has to implement [JENKINS-34036](https://issues.jenkins-ci.org/browse/JENKINS-34036) 
* In general this should improve serverside and clientside performance a bit, especially on masters with frequent job rename/deletion. 
* Turns out to be *highly* useful as a step towards ripping out the flow scanning algorithm (prep for block-scoped stages support).

In the process of fixing server cache invalidation, the server caching is redone to use the FullItemName of the job & its ID - this allows us to just invalidate *specific* run entries upon delete/rename of jobs, rather than invalidating the whole cache.  The bonus is that upon rename we can simply re-key the cache entry and avoid recomputation of the run.  This should reduce load on masters where jobs are often deleted/renamed while viewing.

Unfortunately, this keying change prevented the ability to retrieve data for individual stages from the run data cache, so we added an option to return the full stage data with the run itself (and call this version of the API from the UI clientside).  The clientside caching is also redone in the UI to support this. 

- [x] Add method onDelete to ItemListener to invalidate cache entries when a job is deleted
- [x] Replace the server FlowExecution-url based cache of digested run objects with one based on the WorkflowJob's FullName & the Run ID (this lets you replace cache entries upon rename)
- [x] Option to return full stage data from runs API
- [x] Clientside: use the stage data from the run object, rather than issuing separate requests to get stage node data (greatly reduces requests, and avoids hitting the node APIs which can't use caching).
- [x] Unit tests for the revised rename/delete listeners that are supposed to invalidate the cache.
- [x] Fix UI bug with failed stages & popup

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
@@ -25,17 +25,17 @@
 package com.cloudbees.workflow.flownode;
 
 import com.cloudbees.workflow.rest.external.ExecDuration;
+import com.cloudbees.workflow.rest.external.JobExt;
 import com.cloudbees.workflow.rest.external.RunExt;
 import com.cloudbees.workflow.rest.external.StageNodeExt;
 import com.cloudbees.workflow.rest.external.StatusExt;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-import com.google.common.collect.ImmutableList;
 import hudson.Extension;
-import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 import hudson.model.Item;
 import hudson.model.listeners.ItemListener;
+import hudson.util.RunList;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.actions.NotExecutedNodeAction;
@@ -45,11 +45,11 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.support.actions.PauseAction;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -59,6 +59,7 @@ import java.util.Comparator;
 import java.util.ConcurrentModificationException;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
@@ -114,28 +115,22 @@ public class FlowNodeUtil {
     }
 
     @CheckForNull
-    public static RunExt getCachedRun(FlowExecution ex) {
-        try {
-            if (ex != null) {
-                RunExt cachedRun = CacheExtension.all().get(0).getRunCache().getIfPresent(ex.getUrl());
-                if (cachedRun != null) {
-                    return cachedRun;
-                }
-            }
-            return null;
-        } catch (IOException ioe) {
-            LOGGER.severe("Can't get execution url for execution, IOException!");
-            throw new IllegalStateException(ioe);
+    public static RunExt getCachedRun(@Nonnull WorkflowRun run) {
+        RunExt cachedRun = CacheExtension.all().get(0).getRunCache().getIfPresent(getRunPath(run));
+        if (cachedRun != null) {
+            return cachedRun;
         }
+        return null;
     }
 
-    public static void cacheRun(FlowExecution exec, RunExt run) {
-        if (exec != null && exec.isComplete()) {
-            try {
-                CacheExtension.all().get(0).getRunCache().put(exec.getUrl(), run);
-            } catch (IOException ioe) {
-                LOGGER.severe("Can't get execution url for execution, IOException!");
-            }
+    @Nonnull
+    private static String getRunPath(@Nonnull WorkflowRun run) {
+        return run.getParent().getFullName()+"/"+run.getId();
+    }
+
+    public static void cacheRun(WorkflowRun run, RunExt runExt) {
+        if (!run.isBuilding()) {
+            CacheExtension.all().get(0).getRunCache().put(getRunPath(run), runExt);
         }
     }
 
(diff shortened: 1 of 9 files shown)


svanoort:
@reviewbybees

svanoort:
Note: this opens the door to redoing the API & clientside model, so that the RunExt objects contain all needed data to render the run (ex input actions, changesets, etc).  These extra data elements could be converted to an extension point which decorates the run.   Clientside templating can be simplified as well by working directly with the run data and not doing async callbacks/listeners. 

That change will make it *much* easier to do clientside work for JENKINS-26522, and may reduce load from doing user resolution for changesets (which can sometimes be broken due to SECURITY-243 fixes).

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -114,28 +115,22 @@ private FlowNodeUtil() {
...
-    public static void cacheRun(FlowExecution exec, RunExt run) {
-        if (exec != null && exec.isComplete()) {
-            try {
-                CacheExtension.all().get(0).getRunCache().put(exec.getUrl(), run);
-            } catch (IOException ioe) {
-                LOGGER.severe("Can't get execution url for execution, IOException!");
-            }
+    @Nonnull
+    static String getRunPath(@Nonnull WorkflowRun run) {
+        return run.getParent().getFullName()+"/"+run.getId();
```
jglick: Or could just use

```java
return run.getExternalizableId();
```

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -494,20 +489,67 @@ public static StatusExt getStatus(FlowNode node) {
         return nodes;
     }
 
+    protected static String itemFullNameToUrl(String fullName) {
```
jglick: :ant: apparently unused

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -494,20 +489,67 @@ public static StatusExt getStatus(FlowNode node) {
...
+            // Replace with newFullName
+            CacheExtension ext = CacheExtension.all().get(0);
+            Cache<String, RunExt> rc = ext.getRunCache();
+
             if (item instanceof WorkflowJob) {
-                // It's a cache, we can rebuild it... we have the technology.
-                CacheExtension ext = CacheExtension.all().get(0);
-                ext.getExecutionCache().invalidateAll();
-                ext.getRunCache().invalidateAll();
+                RunList<WorkflowRun> runs = ((WorkflowJob) item).getBuilds().limit(JobExt.MAX_RUNS_PER_JOB+5);  // Add a few to help invalidate just-completed
```
jglick: Could you not just look in the cache for entries with the right key, without actually needing to load any `WorkflowRun`s?

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/flownode/Describe.java
```diff
@@ -52,15 +52,15 @@ public static String getUrl(FlowNode node) {
 
     public static FlowNodeExt get(FlowNode node) {
         if (StageNodeExt.isStageNode(node)) {
-
-            RunExt cachedRunData = FlowNodeUtil.getCachedRun(node.getExecution());
+            // TODO figure out a way to use the cache when we have the node+FlowExectuon but not its parent run
```
jglick: typo

rest-api/src/main/java/com/cloudbees/workflow/rest/external/JobExt.java
```diff
@@ -111,7 +111,12 @@ private static int countRuns(WorkflowJob job) {
     public static List<RunExt> create(List<WorkflowRun> runs) {
         return create(runs, null);
     }
+
     public static List<RunExt> create(List<WorkflowRun> runs, String since) {
+        return create(runs, since, false);
```
jglick: Can deprecate obsolete overloads.

rest-api/src/test/java/com/cloudbees/workflow/flownode/CachingTest.java
```diff
@@ -0,0 +1,114 @@
...
+        Cache<String, RunExt> cache = FlowNodeUtil.CacheExtension.all().get(0).getRunCache();
+
+        // Create one job
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "BlinkinJob");
+        job.setDefinition(new CpsFlowDefinition("" +
+                "stage 'first' \n" +
+                "echo 'done' "
+        ));
+        QueueTaskFuture<WorkflowRun> build = job.scheduleBuild2(0);
+        jenkinsRule.assertBuildStatusSuccess(build);
```
jglick: Even with an old harness, it is more concise to say

```java
WorkflowRun build = jenkinsRule.assertBuildStatusSuccess(job.scheduleBuild2(0));
```

if you are happy to wait for the build to complete before proceeding.

ui/src/main/js/model/job-history-listener.js
```diff
@@ -28,7 +28,7 @@ exports.listen = function (jobUrl, callback) {
     restApi.getJobRuns(jobUrl, function (jobRunsData) {
         callback(jobRunsData);
         setupJobPoll(jobUrl, callback, jobRunsData)
-    });
+    }, {fullStages: 'true'});
```
jglick: Ugly. Why not take a `Boolean` as the type of the `@QueryParameter`?

jglick: Or even `boolean`, since I think it will default to false.

ui/src/main/js/model/rest-api.js
```diff
@@ -28,21 +28,63 @@ var ajax = require('../util/ajax');
  * Workflow REST API
  */
 
+var cache = {};
+
+function addCacheEntry (key, ob) {
+    if (typeof ob === 'object') {
+        if (ob.status && (ob.status === 'SUCCESS' || ob.status === 'ABORTED' || ob.status === 'FAILED' || ob.status === 'UNSTABLE')) {
```
jglick: :bug: why just these, and not, say, `NOT_BUILT` (used for example by `milestone`)?

jglick:
:bee: AFAIUI

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @jglick Not sure what you mean?  It's keyed by the run's ID, not the job -- I could iterate through the entire cache looking for cases with $JobFullName# as a prefix, but the cache has up to 1000 entries.  Is there an API I didn't know about which will simply list IDs of valid builds (in descending order)?

ui/src/main/js/model/rest-api.js
svanoort: @jglick We don't have a NOT_BUILT status code for this (see StatusExt).  Maybe worth caching for NOT_EXECUTED too, but at worst it's a mild performance hit for us.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: Yeah, that's better.  Changing it.

ui/src/main/js/model/job-history-listener.js
svanoort: Had to go digging in stapler + the beanutils converter (ugh) but confirmed that's true.

svanoort:
@jglick Once it's got your re-bee, I'm merging it.

ui/src/main/js/model/rest-api.js
jglick: Sorry, thought this was a `Result`.

jglick:
re:bee:

PR #25, opened 2016-08-16
Use new analysis apis to support block-scoped stages

svanoort:
Implements [JENKINS-34038](https://issues.jenkins-ci.org/browse/JENKINS-34038).
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/10
And also downstream of https://github.com/jenkinsci/pipeline-graph-analysis-plugin

Elements:

- [x] Replace the way we collect stages for a WorkflowRun with an implementation based on a StageFinder (from the pipeline graph analysis plugin) and a SimpleChunkVisitor (from the workflow API extensions)
- [x] Update dependencies and test with block-scoped stage
- [x] Replace the FlowNode Describe API for stages with something that can hit the stage cache (and use the new Chunk visitor) where appropriate. 
- [x] Replace the single-node describe APIs with something that looks just at the node afterward (no need to fetch whole flow). 
- [x] Purge the old flow analysis code with Extreme Prejudice. -- Removed 1000+ lines.
- [x] Manual testing in deployment...


abayer:
:bug: Try https://github.com/abayer/tmp-stage-view-regression-tests/tree/catch-an-error - in old-school stage view, it renders - first stage is successful, second stage fails but continues (as expected with `catchError` which sets the status but doesn't stop the build) and third stage runs fine but is still technically failed because the build status is failure by then. 

But with this, it just doesn't render at all - no stage view for it.

abayer:
:ant: this time 'cos I'm not sure if it's intended or a bug - https://github.com/abayer/tmp-stage-view-regression-tests/tree/parallel-why-not - the error signal in the parallel block doesn't show up in the logs when you click on them, just in the stage view box itself. Is that supposed to be that way?

svanoort:
@abayer I'm working on the first item there, but AFAICT the second one isn't a regression (but it's something I wanted to address as yet-another preexisting issue, a papercut in this case)

svanoort:
@abayer This resolves your error (and includes a test for it), plus pulls in an upstream fix that fixes a simple argument transposition (from https://github.com/jenkinsci/workflow-api-plugin/pull/10/commits/2030360cc8bb6c1a701e92ffec8fa70e613c62c5).

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -73,9 +65,7 @@ private FlowNodeUtil() {
     }
 
     public abstract static class CacheExtensionPoint implements ExtensionPoint {
-        public abstract Cache<String,List<FlowNode>> getExecutionCache();
```
oleg-nenashev: 🐛 Changing of extension point APIs definitely requires 2.0 version of the plugin

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -133,96 +112,43 @@ public static boolean isNotPartOfRunningBuild(FlowExecution execution) {
...
-
-            // Locate the first node in the stage that was actually executed.  This can
-            // vary e.g. when there's was a checkpoint restart, the first executed node
-            // in the stage could be in middle of the stage Pipeline definition.
-            firstExecutedNode = getStageExecStartNode(allNodesSorted, stageStartNodeIndex);
 
-            if (firstExecutedNode != null) {
-                long startTime = TimingAction.getStartTime(firstExecutedNode);
-                long endTime;
+        int iota = Integer.parseInt(node.getId());
```
oleg-nenashev: 🐜 Unhandled NumberFormatException

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/flownode/Describe.java
```diff
@@ -23,48 +23,51 @@
...
+                        }
                     }
                 }
-            }*/
-
-            StageNodeExt stageNodeExt = StageNodeExt.create(node);
-            stageNodeExt.addStageFlowNodes(node);
-            return stageNodeExt;
+            } catch (IOException ioe) {
+                throw new RuntimeException(ioe);
```
oleg-nenashev: 🐛  Looks like a dirty hack. Since you already break binary compatibility, maybe makes sense to perform a nice error propagation. Otherwise - at least document the exception in Javadoc and the code

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -133,96 +112,43 @@ public static boolean isNotPartOfRunningBuild(FlowExecution execution) {
...
-                return execDuration;
+            } catch (IOException ioe2) {
+                // Nope miss again, let's do the harder thing.
             }
         }
 
-        return new ExecDuration();
+        // Find node after this one, scanning everything until this one
+        final FlowNode after = new ForkScanner().findFirstMatch(node.getExecution().getCurrentHeads(), Collections.singletonList(node), new Predicate<FlowNode>() {
+            public boolean apply(@Nonnull FlowNode f) {
```
oleg-nenashev: 🐛 Super method is CheckForNull. Seems there is a missing check

rest-api/src/main/java/com/cloudbees/workflow/rest/external/FlowNodeExt.java
```diff
@@ -154,10 +161,20 @@ public static FlowNodeExt create(FlowNode node, String execNodeName, ExecDuratio
...
+                    WorkflowRun run = (WorkflowRun) exec;
+                    TimingInfo ti = StatusAndTiming.computeChunkTiming(run, PauseAction.getPauseDuration(node), node, node, FlowNodeUtil.getNodeAfter(node));
+                    if (ti != null) {
+                        setStartTimeMillis(ti.getStartTimeMillis());
+                        setPauseDurationMillis(ti.getPauseDurationMillis());
+                        setDurationMillis(ti.getTotalDurationMillis());
+                    }
+                }
+            } catch (IOException ioe) {
+                // Only happens when the WorkflowRun is in an invalid state.  In this case,
```
oleg-nenashev: Incomplete sentence.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/FlowNodeExt.java
```diff
@@ -170,7 +187,7 @@ public void addBasicNodeData(FlowNode node, String execNodeName, ExecDuration du
         set_links(new FlowNodeLinks());
         get_links().initSelf(Describe.getUrl(node));
         setStatus(status);
-        if (status != StatusExt.NOT_EXECUTED) {
+        if (status != StatusExt.NOT_EXECUTED && error != null) {
```
oleg-nenashev: Was it a bug in the previous code?

rest-api/src/main/java/com/cloudbees/workflow/rest/external/FlowNodeExt.java
```diff
@@ -181,8 +198,7 @@ public void addBasicNodeData(FlowNode node, String execNodeName, ExecDuration du
         }
     }
 
-    protected void addBasicNodeData(FlowNode node) {
-        String execNodeName = FlowNodeUtil.getExecNodeName(node);
```
oleg-nenashev: Hmm. Why have you removed it?

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
```diff
@@ -55,23 +57,7 @@ public void setStageFlowNodes(List<AtomFlowNodeExt> stageFlowNodes) {
...
-        // Use the last node in the stage to configure the stage status.
-        FlowNode stageEndNode = FlowNodeUtil.getStageEndNode(node);
-        if (NotExecutedNodeAction.isExecuted(stageEndNode)) {
-            stageNodeExt.setStatus(StatusExt.valueOf(stageEndNode.getError()));
-        } else {
-            stageNodeExt.setStatus(StatusExt.NOT_EXECUTED);
-        }
-
-        return stageNodeExt;
+        return (node.getAction(LabelAction.class) != null && node.getAction(ThreadNameAction.class) == null);
```
oleg-nenashev: IMHO it is unsafe if other nodes somehow add these actions as well. Would it be somehow possible to implement a stronger check?

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
```diff
@@ -103,23 +89,9 @@ public StageNodeExt myWrapper() {
...
-        // node on the stage.
-        FlowNode stageExecStartNode = FlowNodeUtil.getStageExecStartNode(node);
-        if (stageExecStartNode != null) {
-            setStartTimeMillis(TimingAction.getStartTime(stageExecStartNode));
-        }
-        ExecDuration execDuration = FlowNodeUtil.getStageExecDuration(node);
-        setDurationMillis(execDuration.getTotalDurationMillis());
-        setPauseDurationMillis(execDuration.getPauseDurationMillis());
-        setPauseDurationMillis(Math.min(getPauseDurationMillis(), getDurationMillis()));
+        // StageNodeExt objects must have timing passed in
```
oleg-nenashev: Add check and logging for it?
Or IllegalStateException?

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StatusExt.java
```diff
@@ -53,4 +57,22 @@ public static StatusExt valueOf(Throwable t) {
...
+        switch (st) {
+            case PAUSED_PENDING_INPUT: return StatusExt.PAUSED_PENDING_INPUT;
+            case ABORTED: return StatusExt.ABORTED;
+            case FAILURE: return StatusExt.FAILED;
+            case IN_PROGRESS: return StatusExt.IN_PROGRESS;
+            case UNSTABLE: return StatusExt.UNSTABLE;
+            case SUCCESS: return StatusExt.SUCCESS;
+            case NOT_EXECUTED: return StatusExt.NOT_EXECUTED;
+            default:
+                // Shouldn't happen, above includes all statuses
```
oleg-nenashev: Better to have IllegalStateException in order to propagate an error about the bug in the code if somebody adds new state and forgets about this code

rest-api/src/test/java/com/cloudbees/workflow/flownode/FlowNodeUtilTest.java
```diff
@@ -1,208 +0,0 @@
...
-import org.junit.Test;
-
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-
-/**
- * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
- */
-public class FlowNodeUtilTest {
```
oleg-nenashev: 🐛 the class still exists in the code && it has not been even modified by your pull request. Why would you want to delete the entire test class for it?

ui/pom.xml
```diff
@@ -84,6 +70,36 @@
             <version>1.1</version>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>script-security</artifactId>
+            <version>1.18</version>
```
oleg-nenashev: Why do you need explicit dependency? Workflow is supposed to provide it

ui/pom.xml
```diff
@@ -84,6 +70,36 @@
...
+            <scope>test</scope>
+        </dependency>
+        <!--<dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-api</artifactId>
```
oleg-nenashev: dead code

oleg-nenashev:
I am not exactly sure why JENKINS-34038 needs so massive breaking changes. Could you please clarify it?

🐛 for the issues discovered during the code review

rest-api/src/main/java/com/cloudbees/workflow/rest/external/FlowNodeExt.java
svanoort: It's tied to legacy logic and shouldn't be invoked in this fashion directly.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/FlowNodeExt.java
svanoort: Yes.  It was masked by another quirky behavior.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: We're not really following SemVer for plugins or Jenkins as a whole properly anyway, but true. 

This is an extension point *only* to resolve issues with JenkinsRule seeing stale cache data in tests, and has no consumers or implementations but this one, so an :ant: at best.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
svanoort: @oleg-nenashev Which method are you referring to here?

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/flownode/Describe.java
svanoort: @oleg-nenashev It is completely a dirty hack (the way we get at the WorkflowRun), but the only way to do it since pipeline does not provide a better API for it. 

Please take take a closer look at the IOException case here, and I think you'll see why this exception is completely safe to handle in such a fashion -- it's more of a formality handling it at all, since prechecks here should prevent it ever being thrown.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
svanoort: @oleg-nenashev What's your error case here?  We could do a type/descriptor check, but that is far more brittle in the face of subtle implementation changes and this actually lets stage view handle some other labelling schemes if added in the future.  It's also been run by @jglick and at the time he concurred that it was a reasonable approach.

rest-api/src/test/java/com/cloudbees/workflow/flownode/FlowNodeUtilTest.java
svanoort: @oleg-nenashev Why would I keep tests based on a mocking framework with generates an incorrect (mocked) flow graph structure?  Especially when JobAndRunAPI tests cover the same logic in a much deeper functional way?

svanoort:
@oleg-nenashev The algorithms used to extract stages could not be adapted to handle block-scoped stages in anything more than a trivial way, and the architecture prevented reasonable performance.  Let's talk on chat today -- I'll explain all the ways the old algorithms had problems (and how they were so problematic that they could completely lock up Jenkins masters). 

I think you're being overly concerned about back-compatibility here: it's important to know that the Stage View UI is generally the sole consumer of the APIs here.  I'll give you a version of the classic "if a tree falls in the forest" saying: if an API disappears or changes, is it really "breaking" when there's nobody there to complain about it.  ;-)

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
oleg-nenashev: It's a public extension point. So you don't know if there are implementations.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
oleg-nenashev: apply()

abayer:
Ok, revoking my bug with a :bee: since the issue no longer appears.

rest-api/src/test/java/com/cloudbees/workflow/flownode/FlowNodeUtilTest.java
oleg-nenashev: If it is tested elsewhere, I'm fine

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
oleg-nenashev: I would add something like <code>TODO: is it really false-positive-safe?</code> to the code. It seems to be fragile, but I would not expect an issue unless somebody wants to add it intentionally

rest-api/src/main/java/com/cloudbees/workflow/rest/external/FlowNodeExt.java
oleg-nenashev: ok

oleg-nenashev:
>  I'll give you a version of the classic "if a tree falls in the forest" saying: if an API disappears or changes, is it really "breaking" when there's nobody there to complain about it. ;-)

Just bump to 2.0 and I'll be fine

pom.xml
```diff
@@ -46,8 +46,8 @@
     </scm>
 
     <properties>
-        <jenkins.version>1.609.3</jenkins.version>
-        <java.level>6</java.level>
+        <jenkins.version>1.642.3</jenkins.version>
+        <java.level>7</java.level>
         <workflow.version>1.14</workflow.version>
```
jglick: :ant: #7 deletes this; did you really want to keep it?

pom.xml
```diff
@@ -46,8 +46,8 @@
     </scm>
 
     <properties>
-        <jenkins.version>1.609.3</jenkins.version>
-        <java.level>6</java.level>
+        <jenkins.version>1.642.3</jenkins.version>
+        <java.level>7</java.level>
```
jglick: :ant: this is the default, can be elided

rest-api/pom.xml
```diff
@@ -46,39 +46,65 @@
...
-            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-aggregator</artifactId>
+            <version>2.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <!-- Trying this for block-scoped stage support -->
```
jglick: Not sure what this comment is referring to. What block-scoped stage support is there in `workflow-cps`?

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: I think this is fine, though it perhaps should have been `@Restricted(NoExternalUse.class)` from the start.

(Is there _any_ part of this plugin intended to be used as a Java API? I know we make some vague promises about supporting the REST API for other uses. Probably there is some SPI component used to allow proprietary metadata to be used from the UI?)

jglick: FWIW my original recommendation was for there to _not_ be an `ExtensionPoint`, since there was never any expectation of multiple implementations—just a singleton `@Extension`.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: I will treat this as a :bug: since there is no guarantee that IDs are in fact numeric, though in the current implementation they happen to be. (In fact it would be a nice idea to convert them to a dotted-tree notation.)

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -133,96 +112,43 @@ public static boolean isNotPartOfRunningBuild(FlowExecution execution) {
...
+            }
+        } catch (IOException ioe) {
+            try {
+                nextNode = exec.getNode(Integer.toString(iota + 2));
+                if (nextNode != null && nextNode.getParents().contains(node)) {
+                    return nextNode;
                 }
-                return execDuration;
+            } catch (IOException ioe2) {
+                // Nope miss again, let's do the harder thing.
```
jglick: :ant: an `IOException` here (as opposed to a simple null return value) should be logged.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -133,96 +112,43 @@ public static boolean isNotPartOfRunningBuild(FlowExecution execution) {
...
             }
         }
 
-        return new ExecDuration();
+        // Find node after this one, scanning everything until this one
+        final FlowNode after = new ForkScanner().findFirstMatch(node.getExecution().getCurrentHeads(), Collections.singletonList(node), new Predicate<FlowNode>() {
+            public boolean apply(@Nonnull FlowNode f) {
+                List<FlowNode> parents = f.getParents();
+                return (parents != null && parents.contains(node));
```
jglick: :ant: `getParents()` is always non-null (could add such an annotation to one of the upstream PRs)

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: No, the interface method is `@Nullable`, so it is up to the method accepting the `Predicate` to document its nullness semantics. In this case I think the expectation is that the argument should never be null.

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/flownode/Describe.java
jglick: `WorkflowRun.Owner.getExecutable` can quite legitimately throw `IOException`. I do not really care whether you log it, throw it directly out of this method after generalizing its signature, or wrap it in another exception—so long as it is not completely swallowed.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChunkVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+
+        GenericStatus status = StatusAndTiming.computeChunkStatus(run, beforeNode, node, node, next);
+        if (status == null) {
+            status = GenericStatus.NOT_EXECUTED;
+        }
+        ErrorAction err = node.getError();
+        if (status == GenericStatus.FAILURE && err == null) {
+            // Needed for FlowEndNode, since an ErrorAction isn't generated
+            // When build result is set directly
+            err = new ErrorAction(new Throwable("Run failed, assuming last chunk failed too!"));
```
jglick: :ant: IIUC this would become a user-visible message, so its terminology should be chosen more carefully. What is a “chunk” to a user?

Perhaps

```java
new AbortException("Build marked as a failure.");
```

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
jglick: > if added in the future

This will already match some non-`stage` block nodes _today_, but I think that is what we want.

Is this code only called in the case of block nodes? Because it will certainly match plenty of non-block-scoped steps which add `LabelAction` for various reasons, which I think we do _not_ want to detect in this plugin—we only care about (a) legacy `stage` (which adds `LabelAction` and `StageAction`), and (b) block-scoped steps (including but not limited to new `stage`) which add `LabelAction` (but specifically excluding `parallel` which adds `ThreadNameAction`).

jglick: At any rate, :ant: the method name is probably misleading, and there should be some Javadoc explaining whatever it is intended to do.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -52,14 +56,67 @@
...
+        QueueTaskFuture<WorkflowRun> build = job.scheduleBuild2(0);
+        jenkinsRule.assertBuildStatusSuccess(build);
+
+        JenkinsRule.WebClient webClient = jenkinsRule.createWebClient();
+
+        assertBasicJobInfoOkay(job, webClient);
+        assertArtifactsEndpointOkay(job, webClient);
+        assertChangesetsEndpointOkay(job, webClient);
+
+        /** Verify the runs show right stages as per {@link #assertBasicRunInfoOkay(WorkflowJob, JenkinsRule.WebClient)} */
```
jglick: BTW Javadoc is impossible between statements like this.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -70,10 +127,8 @@ public void test() throws Exception {
                 "   stage ('Test'); " +
                 "   echo ('Testing'); " +
                 "   stage ('Deploy'); " +
-                (hudson.Functions.isWindows() ? "   bat ('rmdir /s/q targs || echo no such dir\\n mkdir targs && echo hello> targs\\\\hello.txt'); "
-                        : "   sh ('rm -rf targs && mkdir targs && echo hello > targs/hello.txt'); "
-                ) +
-                "   archive(includes: 'targs/hello.txt'); " +
+                "     writeFile file: 'file.txt', text:'content'; " +
+                "     archive(includes: 'file.txt'); " +
```
jglick: more concisely:

```groovy
archive 'file.txt'
```

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -95,6 +150,58 @@ public void test() throws Exception {
...
+
+        Assert.assertEquals(StatusExt.SUCCESS, runExt.getStages().get(0).getStatus());
+        Assert.assertEquals(StatusExt.SUCCESS, runExt.getStages().get(1).getStatus());
+        Assert.assertEquals(StatusExt.FAILED, runExt.getStages().get(2).getStatus());
+
+        StageNodeExt finalStage = runExt.getStages().get(2);
+        AtomFlowNodeExt finalNode = finalStage.getStageFlowNodes().get(finalStage.getStageFlowNodes().size()-1);
+        Assert.assertEquals(StatusExt.FAILED, finalNode.getStatus());
+        Assert.assertNotNull(finalNode.getError());
+        Assert.assertNotNull(finalNode.getError().getMessage());
```
jglick: If you localize the placeholder error message using `Messages` you could `assertEquals` here.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -129,10 +231,10 @@ private void assertBasicJobInfoOkay(WorkflowJob job, JenkinsRule.WebClient webCl
         JSONReadWrite jsonReadWrite = new JSONReadWrite();
         JobExt jobExt = jsonReadWrite.fromString(jsonResponse, JobExt.class);
 
-        Assert.assertEquals("Noddy Job", jobExt.getName());
+        Assert.assertEquals(job.getName(), jobExt.getName());
         Assert.assertEquals(1, jobExt.getRunCount());
-        Assert.assertEquals("/jenkins/job/Noddy%20Job/wfapi/describe", jobExt.get_links().self.href);
-        Assert.assertEquals("/jenkins/job/Noddy%20Job/wfapi/runs", jobExt.get_links().getRuns().href);
+        Assert.assertEquals("/jenkins/" + job.getUrl() + "wfapi/describe", jobExt.get_links().self.href);
```
jglick: `/jenkins` here is `jenkinsRule.contextPath`

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -161,34 +263,81 @@ private void assertDescribeEndpointOkay(WorkflowJob job, JenkinsRule.WebClient w
...
+            Assert.assertTrue(childNode.getId() != null && !(childNode.getId().isEmpty()));
+            // Handle me: errorExt
+            ErrorExt err = childNode.getError();
+            if (err != null) {
+                Assert.assertNotNull(err.getMessage());
+                Assert.assertNotNull(err.getType());
+            }
+            Assert.assertNotNull(childNode.getStatus());
+
+            Assert.assertTrue("Start times should always be > 0 after 1970", childNode.getStartTimeMillis() > 0);
```
jglick: If you remember block-scoped stages in the 1960s, you weren’t there!

ui/pom.xml
jglick: From #7; sometimes you need to override an overly old dependency from the wrong path in the tree. (Whether the above is actually new enough, I am not sure.)

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
jglick: In fact it seems you _did_ add such an annotation, so I wonder why FindBugs is not complaining here about an unnecessary null check.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
svanoort: It won't generate true Javadocs, but it points someone helpfully at the right place (and IIRC IntelliJ is sometimes smart enough to do links inside comments anyway).

jglick:
:bee:

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
svanoort: Dropping in to mention: I actually want it to return "false positives" -- this gives some measure of forward compatibility to stage view here, and lets us display additional labelled blocks.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -172,6 +183,83 @@ public static StatusExt getStatus(FlowNode node) {
...
+
+
+    /** List the stage nodes for a FlowExecution -- needed for back-compat.
+     *  Note: far more efficient than existing implementation because it uses the cache of run info.
+     */
+    @Nonnull
+    public static List<FlowNode> getStageNodes(@CheckForNull FlowExecution execution) throws RuntimeException {
+        WorkflowRun run = getWorkflowRunForExecution(execution);
+        if (run == null) {
+            return Collections.EMPTY_LIST;
```
jglick: :ant: use `emptyList()` not `EMPTY_LIST`

jglick:
re:bee:

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/flownode/Describe.java
oleg-nenashev: I'm still not fine with it since it's not documented. Keeping 🐛

rest-api/src/main/java/com/cloudbees/workflow/rest/external/FlowNodeExt.java
oleg-nenashev: Not addressed

oleg-nenashev:
Keeping 🐛 for undocumented RuntimeException, which is being explicitly thrown for a plausible case.

After quick re-check I do not see other significant issues.

rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/flownode/Describe.java
jglick: I do not think documentation is particularly relevant here, since the interesting API is via REST, not Java. But at any rate, the simplest thing would be to just add `throws IOException` and anything calling it; ultimately you are up to web methods, and who cares.

jglick: Note that whether `Owner.getExecutable` would _actually_ throw an exception under these circumstances is hard to say without deeper analysis. Typically the `Owner.run` is set either when a build starts, or is resumed, and then is cached for the rest of the session, in which case the method returns instantly with no exception. On the other hand I have had reports of cases where this is thrown for reasons I do not understand—not known how to reproduce.

Given that the ultimate caller is going to be a JavaScript client, and that there is nothing much better we can do with this error besides log it with any other error and give up, wrapping in a `RuntimeException` seems no worse than anything else.

oleg-nenashev: If it gets documented, I'll accept it.

> Given that the ultimate caller is going to be a JavaScript client

It's a public API. We do not know who calls it

oleg-nenashev:
🐝

pom.xml
```diff
@@ -33,7 +33,7 @@
 
     <groupId>org.jenkins-ci.plugins.pipeline-stage-view</groupId>
     <artifactId>parent-pom</artifactId>
-    <version>2.0-SNAPSHOT</version>
+    <version>2.0-beta-2-SNAPSHOT</version>
```
jglick: Do not forget to revert…

pom.xml
```diff
@@ -48,7 +48,7 @@
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
         <java.level>7</java.level>
-        <workflow.version>1.14</workflow.version>
+        <workflow.version>2.0</workflow.version>
```
jglick: :ant: this [does not make any sense](https://jenkins.io/blog/2016/04/22/pipeline-2/); specify the version of each plugin separately

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChunkVisitor.java
```diff
@@ -64,9 +65,16 @@ public static AtomFlowNodeExt makeAtomNode(@Nonnull WorkflowRun run, @CheckForNu
 
     @Override
     /** Do the final computations to materialize the stage */
+    @SuppressFBWarnings(value = "RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE", justification = "We can actually get nulls")
```
jglick: What could actually be null that FB thinks cannot? Would be worth explaining in the `justification`.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChunkVisitor.java
```diff
@@ -96,13 +105,18 @@ protected void resetChunk(@Nonnull MemoryFlowChunk chunk) {
         super.resetChunk(chunk);
         firstExecuted = null;
         stageNodeIds.clear();
+        stageContents.clear();
     }
 
     @Override
+    @SuppressFBWarnings(value = "RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE", justification = "We can actually get nulls")
```
jglick: Ditto.

jglick:
re:bee:

rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChunkVisitor.java
svanoort: FB wasn't inspecting deeply enough -- there was a case it misses.   This is the tradeoff for expecting fast testing.

svanoort:
Merging, with caveat that I may open up another PR for any smaller bugs from integration testing, and will have to do a tiny one for dependencies.

pom.xml
svanoort: Will be done in a final PR before release.

PR #26, opened 2016-08-26
POM fixes for final 2.0 release

svanoort:
Includes a whole raft of small POM updates/cleanups as we move the block-scoped stage release train along. 

Based on top of https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/25


jglick:
I suspect this PR is botched and needs to be discarded.

svanoort:
@jglick nah, but it got merged into the main PR so I can close it.

PR #27, opened 2016-08-29
Take in final pom versions and bump version for release

svanoort:
What is says on the tin.

Picks up https://github.com/jenkinsci/pipeline-graph-analysis-plugin/pull/5

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
 
     <groupId>org.jenkins-ci.plugins.pipeline-stage-view</groupId>
     <artifactId>parent-pom</artifactId>
-    <version>2.0-beta-2-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


svanoort:
Comment: will fail to build until block scoped stage is released fully to the public.

jglick:
> will fail to build until block scoped stage is released fully to the public

So this is not technically ready for review.

For things like this I think no PR is needed—you are just bumping up dependencies to public releases, and setting the next version.

svanoort:
@jglick Stage view and block-scoped stage should be released at the same time, to ensure compatibility for all users.

svanoort:
Merging since it's just a pom update and changes are already validated

PR #32, opened 2016-10-02
Empty white space on screens with high resolutions

recena:
@reviewbybees specially @svanoort 

### Before
oss-1473_before

### After
oss-1473_after

### DIFF
--- a/ui/src/main/js/view/templates/pipeline-staged.hbs
+++ b/ui/src/main/js/view/templates/pipeline-staged.hbs
@@ -1,7 +1,7 @@
 <div id="pipeline-box">
 <h2>{{fragCaption}}</h2>
 {{#if numStages}}
-<div class="table-box"><div class="table-viewPort" style="max-width:{{maxTableEms}}em">
+<div class="table-box"><div class="table-viewPort">
 <table class="jobsTable" style="max-width:{{maxTableEms}}em">
   <colgroup>
     <col class="start-group" />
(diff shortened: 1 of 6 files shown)


svanoort:
My system (chrome 52, mac os X retina):  

Jenkins 2.7.19.x (initially with stage view 1.6):

**Before fix:**

screen shot 2016-10-02 at 3 27 13 pm

**After fix:**
------- pending a build -----

ui/src/main/js/view/templates/pipeline-staged.hbs
```diff
@@ -1,7 +1,7 @@
 <div id="pipeline-box">
 <h2>{{fragCaption}}</h2>
 {{#if numStages}}
-<div class="table-box"><div class="table-viewPort" style="max-width:{{maxTableEms}}em">
+<div class="table-box"><div class="table-viewPort">
 <table class="jobsTable" style="max-width:{{maxTableEms}}em">
```
svanoort: This is actually set here: https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/master/ui/src/main/js/view/pipeline-staged.js#L61

But in truth, I think it may not be needed -- seems to play fine with very wide tables (you simply scroll right).

svanoort:
🐝  from review and testing in a variety of scenarios in deployment.

ui/src/main/js/view/templates/pipeline-staged.hbs
recena: It is still needed for `jobsTable`

stephenc:
:bee: from the 🕶  on 🏇

PR #35, opened 2016-11-23
Do not throw the stage view under the test trend graphs (Checkstyle, …

marius-balteanu:
…PMD, ...)

I'm using Jenkins 2.19.3 with the latest versions for pipelines and static analysis collector plugins. Starting with [1.49](https://wiki.jenkins-ci.org/display/JENKINS/Analysis+Collector+Plugin), the graphs are shown in the pipeline view, but because of that, the pipelines are under these graphs even if there is enough space.

Screenshot (Chrome 54.0.2840.99)
image

Removing the "clear: right" seems to solve this issue with no side effects from my basic checks.

image

### DIFF
--- a/ui/src/main/less/stageview_adjunct.less
+++ b/ui/src/main/less/stageview_adjunct.less
@@ -1,6 +1,5 @@
 // The main container for CBWF Stage View
 .cbwf-stage-view {
-  clear: right;
   padding: 15px 0;
 }
 


tknerr:
looks much better than before :)

mbidewell:
I agree, this change make much better use of space.

svanoort:
Sensible to me too.

tlewis:
Yeah, the test trend graph links cannot be interacted with anymore. Can this be tweaked to make them usable again?

PR #34, opened 2016-11-14
[FIX JENKINS-34538] Fix duplicated rows when displayName is changed

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

PR #36, opened 2016-11-29
Fix incorrect API call from frontend which fetched all runs even if they were cached [JENKINS-40087]

svanoort:
Solves [JENKINS-40087](https://issues.jenkins-ci.org/browse/JENKINS-40087) -- root cause is that the "since" parameter to the REST API takes the name of the run ("#3" for example) rather than its id ("3"). 

Thus we were repeatedly fetching all the runs when this was changed by https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/34.

@reviewbybees

### DIFF
--- a/ui/src/main/js/model/job-history-listener.js
+++ b/ui/src/main/js/model/job-history-listener.js
@@ -39,17 +39,17 @@ exports.schedulePoll = function (callback) {
 function setupJobPoll(jobUrl, callback, jobRunsData) {
     function findSinceRunParam() {
         // console.log('findSinceRunParam');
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
🐝 after going through https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/34
Maybe deserves some tests, but generally it's just a revert operation

svanoort:
As @oleg-nenashev says, indeed.  This isn't quite as efficient/precise as querying runs by ID directly but retains most of the benefits of the original fix without needing too much API rework/creation.

PR #37, opened 2017-01-10
[JENKINS-36543] Honor defaultValue=false in BooleanParameterDefinition

jbq:
Currently there is no way to make a checkbox unchecked by default because of a bug in `run-input-required.hbs`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.9</version>
+        <version>2.22</version>
     </parent>
 
     <groupId>org.jenkins-ci.plugins.pipeline-stage-view</groupId>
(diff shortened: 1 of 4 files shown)


jbq:
Hi @svanoort I added some unit tests for this PR.  Any chance you can review it?

Note that I also had to update the parent POM to let the build pass.  I don't know if that change is likely to be integrated?

Quorce:
Any update? When will PR be merged and released?

svanoort:
Close/re-open to retrigger CI build.  As far as the review goes, this looks reasonable to me.

svanoort:
I'm not sure why the CI build is being finnicky, but I've verified it runs cleanly locally, and merging.  

Thank you for the contribution, and apologies on the long delay before being able to review it properly.

pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.9</version>
+        <version>2.22</version>
```
jglick: I think this broke the build due to https://github.com/jenkinsci/plugin-pom/pull/36. Missed this one during my scan, probably because `ui/` is a subdirectory.

jglick: cf. #42

svanoort: Indeed, and it wasn't visible with normal local builds because I had that content cached, and looked like just another transitory environmental issue with CI.  Solved now though.

PR #38, opened 2017-03-02
Fix finding the end of the fhinal chunk in the presence of incomplete parallel branches [JENKINS-38536]

svanoort:
Fixes [JENKINS-38536](https://issues.jenkins-ci.org/browse/JENKINS-38536) in combination with the work in https://github.com/jenkinsci/workflow-api-plugin/pull/33

### DIFF
--- a/rest-api/pom.xml
+++ b/rest-api/pom.xml
@@ -47,7 +47,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.4</version> <!-- allows consuming the new APIs -->
+            <!-- Temp version until final is released -->
+            <version>2.12-alpha</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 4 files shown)


svanoort:
WIP while I sort out why the tests are failing.

svanoort:
@reviewbybees please?

rest-api/pom.xml
```diff
@@ -47,7 +47,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.4</version> <!-- allows consuming the new APIs -->
+            <!-- Temp version until final is released -->
+            <version>2.12-alpha</version>
```
jglick: Be super careful with `alpha`/`beta` dependencies. `maven-release-plugin` will let you go ahead and release this plugin as is—but if anyone actually tries to install it from the non-experimental update center, Jenkins will happily let you go ahead, and then issue an error when you restart. Oops.

So I tend to say that you should only use (timestamped) `SNAPSHOT` dependencies until there is a full upstream release.

svanoort:
@reviewbybees  done

PR #40, opened 2017-05-04
Make build title link to the build itself

martinmine:
Currently, to view a particular build from the pipeline view, you have to find the corresponding job in the build list. To make things more convenient, make the build title link to the build itself, just as each build is in the build list. 
Here is an example how it looks like:


svanoort:
Oh, I like this, and you included a screenshot too -- excellent!  

Need to do some manual testing to cover different browsers and screen resolutions (might be a little while before there's time) but if everything looks good, it'll get merged in.

svanoort:
@martinmine  Thank you for the contribution!  I was just able to loop back to test, and it looks good to me, so merging this and it'll be in the next release.

PR #42, opened 2017-05-25
Fix build broken by changes to node use in new parent pom

svanoort:
Similarly to https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/37/files#r118344817 so see https://github.com/jenkinsci/workflow-cps-plugin/pull/87

Thanks to @jglick  for pointing me at the culprit for the latest failure.

It's possible that some of the frontend plugin dependencies might be removable -- will tinker with that. 

Manually tested to ensure it will still work for users in the wild, not just testing & HPI:run (which handle adjuncts a bit differently).

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -8,6 +8,10 @@
     "moment": "~2.8.3",
     "window-handle": "^1.0.0"
   },
+  "scripts": {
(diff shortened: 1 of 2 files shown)


svanoort:
@reviewbybees

PR #43, opened 2017-05-30
Fix broken caching handling of job renames by purging cache entry [JENKINS-38159]

svanoort:
[JENKINS-38159](https://issues.jenkins-ci.org/browse/JENKINS-38159) - log URLs would have to be rewritten (as well as all HAL links in JSON), so it's better to just purge the cache entry and rebuild when needed.

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
@@ -24,7 +24,6 @@
  */
 package com.cloudbees.workflow.flownode;
 
-import com.cloudbees.workflow.rest.external.JobExt;
 import com.cloudbees.workflow.rest.external.RunExt;
 import com.cloudbees.workflow.rest.external.StageNodeExt;
 import com.cloudbees.workflow.rest.external.StatusExt;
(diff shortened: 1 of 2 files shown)


svanoort:
@reviewbybees

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -267,36 +267,33 @@ public static WorkflowRun getWorkflowRunForExecution(@CheckForNull FlowExecution
...
-                    if (!r.isBuilding()) {
-                        String path = oldFullName+"#"+r.getId();
-                        RunExt cachedRun = rc.getIfPresent(path);
-                        if (cachedRun != null) {
-                            rc.invalidate(path);
-                            rc.put(newFullName+"#"+r.getId(), cachedRun);
-                        }
-                    }
+            ConcurrentMap<String, RunExt> runMap = rc.asMap();
+            for (String cacheRunId : runMap.keySet()) {  // Put the Concurrent in ConcurrentMap to work for us
```
jglick: Maybe better style to use `Iterator.remove`.

rest-api/src/main/java/com/cloudbees/workflow/flownode/FlowNodeUtil.java
```diff
@@ -267,36 +267,33 @@ public static WorkflowRun getWorkflowRunForExecution(@CheckForNull FlowExecution
...
+                // Null-check may not be needed, but just in case of mutation by another thread
```
jglick: Would that really result in null entries??

svanoort: @jglick  I hope not, but am feeling a bit cautious - plus it's easier to put the check than to prove it safe.

PR #44, opened 2017-06-08
Fix run timing bug for runs [JENKINS-40162]

svanoort:
Also incidentally fix computation of queue durations (legacy logic was not using run startTime/timestamp like it should, apparently un-noticed because the difference is small).

Run duration (as separate from stage) was not getting the end time directly from the run like it ought to. 

[JENKINS-40162](https://issues.jenkins-ci.org/browse/JENKINS-40162).

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
@@ -189,7 +189,10 @@ public class RunExt {
 
     /** Computes timings after the stages have been set up
      *  That means timing of the stage has been computed, and the stages are sorted
+     *
+     *  Deprecated but retained for external APIs consuming it, use {@link #createNew(WorkflowRun)} instead
      */
+    @Deprecated
     public static RunExt computeTimings(RunExt runExt) {
 
         // Pause is the sum of stage pauses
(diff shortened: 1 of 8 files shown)


svanoort:
@reviewbybees

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
```diff
@@ -189,7 +189,10 @@ public void setArtifacts(Link artifacts) {
 
     /** Computes timings after the stages have been set up
      *  That means timing of the stage has been computed, and the stages are sorted
+     *
+     *  Deprecated but retained for external APIs consuming it, use {@link #createNew(WorkflowRun)} instead
```
jglick: Wait, there is a user-visible bug and the only non-test, non-formatting change is in a method which you are also deprecating? Does not seem to make sense.

Did you instead mean to mark with `@Restricted(NoExternalUse.class)`, i.e., this method is still used in production but was not intended to be called from outside of this plugin and was only accidentally `public`?

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
```diff
@@ -281,7 +332,28 @@ private void assertBasicRunInfoOkay(WorkflowJob job, JenkinsRule.WebClient webCl
...
+            for (StageNodeExt stage : runExt.getStages()) {
+                assertStageInfoOkay(stage, testStageChildren);
+                stageRunTime += stage.getDurationMillis();
+                stagePauseTime += stage.getPauseDurationMillis();
+            }
+        }
+        Assert.assertTrue("Run has zero duration!", runExt.getDurationMillis() > 0);
+        Assert.assertTrue(String.format("Run duration (%d) ms is shorter than sum of stage runtimes (%d ms), this is insane!",
+                    runExt.getDurationMillis(), stageRunTime),
+                runExt.getDurationMillis() >= stageRunTime);
```
jglick: Should these just be invariants `assert`ed in the `main` code?

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/UndefinedWorkflowTest.java
```diff
@@ -65,8 +65,8 @@ public void test_success_flow() throws Exception {
 
     private void assertRunOkay(RunExt workflowRun) {
         Assert.assertEquals(0, workflowRun.getStages().size());
-        Assert.assertEquals(0L, workflowRun.getEndTimeMillis());
+        Assert.assertTrue("Run that should be near instant wasn't", workflowRun.getDurationMillis() < 100);
```
jglick: Sounds like it will be flaky.

rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
svanoort: @jglick I can understand the confusion -- this is a case where the diff display is being completely horrible.  I had to expand all the sections to confirm something hadn't gone amiss, but the actual logic changes are in `createNew` (and Github "conveniently" collapses the end of `computeTimings` and the beginning of `createNew`).

It's clearer if you check it out and pull it up in meld or another diff tool. 

You're right that this should not be public.  But as far as `@Restricted` -- my concern is that someone's plugin *might* be invoking this in another plugin (library API, not REST) -- which that annotation would break. 

Tempting to just remove the method and risk it though, since consumers are limited and someone *should not* be using this method anyway.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/JobAndRunAPITest.java
svanoort: Absolutely not - then an assertion failure will break the main method, and the few cases where it would happen indicate something anomalous like a clock time changing.

rest-api/src/test/java/com/cloudbees/workflow/rest/endpoints/UndefinedWorkflowTest.java
svanoort: Possible, yes.  If that happens, we can increase the timeout (but if it takes >100 millis there's probably something else wrong).

PR #46, opened 2017-08-12
Display empty stages with another color and status

vvanholl:
When using declarative pipelines, skipped stages from a "when" clause are
displayed as "SUCCESS" in green. This is a bit misunderstanding because
you can think the stage was effectively ran.

With this commit, skipped stages are displayed in gray and have the
"EMPTY" status.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,1 @@
+buildPlugin()
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChunkVisitor.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/ChunkVisitor.java
@@ -83,18 +83,22 @@ public class ChunkVisitor extends StandardChunkVisitor {
         }
         ExecDuration dur = (times == null) ? new ExecDuration() : new ExecDuration(times);
 
-        GenericStatus status;
+        StatusExt status;
         long startTime = 0;
         if (firstExecuted == null) {
-            status = GenericStatus.NOT_EXECUTED;
+            status = StatusExt.fromGenericStatus(GenericStatus.NOT_EXECUTED);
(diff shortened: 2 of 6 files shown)


vvanholl:
As an exemple with the following pipeline :

````
pipeline{
    agent any
    stages{
        stage("build"){
            steps{
                sh("echo build")
            }
        }
        stage("test"){
            steps{
                sh("echo test")
            }
        }
        stage("publish"){
            when{
                branch "master"
            }
            steps{
                sh("echo publish")
            }
        }
    }
}
```
![image](https://user-images.githubusercontent.com/2646864/29242371-67c4039a-7f8c-11e7-9415-f492785d540d.png)

Sulinskii:
Hey, good job! I find it very useful for me.

svanoort:
@vvanholl This looks like a pretty reasonable addition and AFAICT the code looks correct (though it's generally preferred to see a new status in the Pipeline Graph Analysis plugin, where Blue Ocean can consume it, if possible).  

However, it does still need testcases before I can merge it.

mvanini:
Good idea! Waiting for the merge.
I think it would be more clear if you change "empty" with "skipped".

svanoort:
@mvanini It's not mergeable without appropriate test coverage -- if you are waiting for merge, you can make that happen by providing a PR replacing this one which uses this branch as its base and includes tests.

alexmc6:
I agree this is a good idea. I guess the original requester has lost interest in this PR? 
I would be interested to see if this appears in historical builds where a stage has been introduced. Currently those stages disappear - when IMHO it would be better for them to still be displayed - but with the missing stages in grey.

anandvarkeyphilips:
Is it solved now??

dcalap:
Any advance or workaround/plugin?

PR #49, opened 2017-09-27
[JENKINS-43826] Exclude skipped stages from estimated duration calculation

Sulinskii:
This changes allows to exclude skipped stages from estimated duration calculation.
Details are here https://issues.jenkins-ci.org/browse/JENKINS-43826


MPieter:
@svanoort  Any idea of this will be merged? 
We are experiencing here a similar issue.

PR #50, opened 2017-10-08
Fixed grammar mistake (1 commits -> 1 commit)

ChadiEM:
It bothers me that I have to repeat the same change in several locations. Any hints on how this can be centralized?


svanoort:
Thanks for submitting this fix!    I think the template use could be refactored but it is probably not worth it when they're so small.

PR #55, opened 2018-06-14
[JENKINS-46096] Avoid display at bottom of page

ssbarnea:
Fixes [JENKINS-46096](https://issues.jenkins-ci.org/browse/JENKINS-46096)

### DIFF
--- a/ui/src/main/less/stageview_adjunct.less
+++ b/ui/src/main/less/stageview_adjunct.less
@@ -1,5 +1,6 @@
 // The main container for CBWF Stage View
 .cbwf-stage-view {
+  clear: right; // see JENKINS-46096
   padding: 15px 0;
 }
 


ssbarnea:
@svanoort Please review this oneliner, which reverts https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/35

svanoort:
@ssbarnea Thanks for the contribution, but I think a simple revert won't do it -- the idea behind #35 was solid *if* there was enough screen real-estate -- I think it needs a more comprehensive fix so the layout reflows correctly. 

For review of the CSS change, we should also have a before-and-after screenshots to show the bug and the result of the new layout.   Thanks!

JeanChristopheMorinRodeoFX:
Hi guys, just wondering if there was any work being done on this? Having the stage view overflowing on the warnings/tests report is a real pain... I think I'd personally prefer to have the stage view lower in the screen than having a broken UI. Yes it's not pretty to the eyes but at least all widgets/components would work compared to what we have right now.

What do you guys think?

TobiX:
This is now hopefully obsolete, thanks to #72

PR #58, opened 2018-08-27
get input submitter from rest api

runzexia:


runzexia:
@svanoort Could you please help review this? Any comments is appreciated. Thanks!

alexanderrtaylor:
@runzexia This is actually already grabbed by the input step as a groovy variable. Are you just looking to be able to also grab some other way? I am confused by the purpose behind this (sorry for being late)

PR #60, opened 2018-10-04
Migrate into localization-zh-cn-plugin

LinuxSuRen:
See https://github.com/jenkinsci/localization-zh-cn-plugin


LinuxSuRen:
Ok

oleg-nenashev:
Still makes sense to have labels so that we track that on our side

PR #61, opened 2018-10-04
Link to full node log from the window

olivergondza:
Switching from the convenient log thing to full log listing is annoying, and in case the logs are longish, often necessary. This adds direct link.

psw-link


olivergondza:
The build failure appears to the the same windows problem master have.

olivergondza:
@svanoort, can you have a look, please?

olivergondza:
Let's see if there are opinions elsewhere - @jenkinsci/code-reviewers

svanoort:
@olivergondza  I'm seeing that tests are still showing failures here -- please don't simply merge directly.

PR #62, opened 2018-11-12
Improve usability of the download artefacts popup 

VirtualTim:
For moderately long artefact names the popup takes up too much vertical space. To me it makes sense to try and keep longer names on one line. I've also increased the area of the button (no visual change) so the popup is less likely to disappear when a user tries to mouse over it.

This should prevent the list looking like this:
image

And make it look like this:
image


VirtualTim:
Also with a large number of artefacts with long names the list looks a bit silly:
image

VirtualTim:
So, this PR has been open a few months, any thoughts on this? The changes are pretty small.

PR #63, opened 2018-11-19
Stabilize order of stages

Nuh:
Add bugfix for duplication of stages


jsoref:
This sounds useful, but in general code changes ask for testcases, can you provide one?

lsouchard:
Is this pull request still active? I really need stabilization of stage orders and it seems like the build failed only because equals and hashCode functions of StageNodeExt have not been overriden. I can provide test cases. If this pull request is not active anymore I can create my own but I'd rather not steal the author's commit.

Nuh:
Done in #88, Thanks @lsouchard!

PR #64, opened 2018-12-12
Revert "Link to full node log from the window"

svanoort:
Reverts jenkinsci/pipeline-stage-view-plugin#61


svanoort:
Original PR was merged without any reviews.

PR #66, opened 2019-02-19
[JENKINS-51490] Caption for Abort Button

benpdavison:
For ticket [JENKINS-51490](https://issues.jenkins-ci.org/browse/JENKINS-51490)

I have replicated the work done to caption the Ok button for the abort button.

This is a something that I often see requested.

Please let me know if there is anything I have missed, or if i need to write a test. I noticed there was no test for the Ok button, so also followed that path.

I currently have a PR in place for the input step plugin to support this logic here: https://github.com/jenkinsci/pipeline-input-step-plugin/pull/32

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/PendingInputActionsExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/PendingInputActionsExt.java
@@ -43,6 +43,7 @@ public class PendingInputActionsExt {
     private String id;
     @JsonInclude(JsonInclude.Include.NON_NULL)
     private String proceedText;
+    private String abortText;
     private String message;
     private List<InputParameterDefExt> inputs;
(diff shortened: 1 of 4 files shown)


rest-api/src/main/java/com/cloudbees/workflow/rest/external/PendingInputActionsExt.java
```diff
@@ -111,6 +114,7 @@ public static PendingInputActionsExt create(InputStepExecution inputStepExecutio
         String inputId = inputStepExecution.getId();
         inputActionExt.setId(inputId);
         inputActionExt.setProceedText(inputStepExecution.getInput().getOk());
+        inputActionExt.setAbortText(inputStepExecution.getInput().getAbort());
```
alexanderrtaylor, line 117: I think you meant:

```suggestion
        inputActionExt.setAbortText(inputStepExecution.getInput().getAbortText());
```

getAbort() is not a method

PR #67, opened 2019-02-28
 Fix overlap when used with Test Result Trend

philipwhiuk:
Without clear, the pipeline stage-view overlaps the Test Result Trend graph that is floated to the right.


philipwhiuk:
@svanoort it looks like the tests fail on HEAD for Windows due to line endings... Care to merge-squash this anyway?

philipwhiuk:
Example of the problem without the clear
overlap - copy

PR #53, opened 2018-04-24
Avoid generating NaNy NaNd

jsoref:
We have a stage which doesn't happen...

Declarative: Checkout SCM | SCM clean up (avoid conflicts) | npm install | npm test | npm publish | Results | clean up
-- | -- | -- | -- | -- | -- | --
| 3s | NaNy NaNd | 30s | 40s | 1min 23s | 1s | 2s

It's been like this for a long time, the `NaNy NaNd` is pretty ugly.

PR #68, opened 2019-05-31
use standard isNaN function

brainsam:
Number.isNaN() defined in ECMAScript 2015 (6th Edition, ECMA-262), that is not supported by IE11.
To provide compatibility, we have to use isNaN function

https://www.w3schools.com/js/js_versions.asp
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#Specifications


dwnusbaum:
Linking to the PR that added the call to `Number.isNaN` for future reference: https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/53.

PR #72, opened 2019-10-12
JENKINS-46096 display the stage view as inline-block

timorohwedder:
To avoid overlay of stage view over test result and other graphs and to
not display the stage swim lanes on the bottom of the page.


TobiX:
I feel this might be the better fix then #67, which creates quite a lot of white space in our Jenkins installation (many trend graphs, so the change in #67 pushes the stage view far to the bottom...)

TobiX:
As a minor nit: `clear: right` now seems useless, the whole pipeline stage view is reflowed beside the trend graphs, in spite of its existence.

Roemer:
Any plans when there will be a release with that fix?

PR #77, opened 2019-12-02
Fix formatter.time with millis value `undefined` returning "NaNy NaNd"

joelcox22:
Fix time formatter returning `NaNy NaNd` when `millis` is undefined.

This happens when all visible builds have skipped a stage, and ends up looking like this:

image


mattcolman:
oh YES YES YES!! ^^

joelcox22:
Never mind :trollface: 

We just need to update the plugin to latest which doesn't use `Number.isNaN`

PR #78, opened 2019-12-04
Update hpi plugin to remove groovy-sandbox.jar from WEB-INF/lib and add license information in pom file

angry-cellophane:
Hi,

In short, two changes:
1. Plugin-pom updated to the latest version (3.54) to use the latest version of the hpi plugin with the fix of https://issues.jenkins-ci.org/browse/JENKINS-58771. Without it the pipeline-stage-view-plugin
hpi file has groovy-sandbox-1.10.jar in the WEB-INF/lib folder that is loaded at runtime. Version 1.10 is quite old and has security vulnerabilities. Besides, it shouldn't be bundled in the first place as it comes as a transitive dependency of a test dependency.
2. Added licence tags in the pom file. 

More about the changes:
1. groovy-sandbox-1.10.jar is bundled in the hpi file due to this bug - https://issues.jenkins-ci.org/browse/JENKINS-58771
```
» ls rest-api/target/pipeline-rest-api/WEB-INF/lib 
groovy-sandbox-1.10.jar  pipeline-rest-api.jar

```

Dependency tree:
```
[INFO] +- org.jenkins-ci.plugins.workflow:workflow-cps:jar:2.31:test
[INFO] |  +- org.jenkins-ci.plugins:script-security:jar:1.26:compile
[INFO] |  |  \- org.kohsuke:groovy-sandbox:jar:1.10:compile
```

2.  Licence tags in the pom file
I'm writing a jenkins plugin for some organization and in order to use this plugin as a dependency it should be validated first. One of the validation is checking the license. There is some automation around it (using [snyk](https://snyk.io/)). Currently, the license is defined in the pom file as a comment, which is not parsed by the automated checker. 
I thought raising a PR to update the license section in the pom file can be quicker than onboarding it manually. Additionally, it can help other rare enterprise developers like me writing custom Jenkins plugins to deal with such bureaucracy.


angry-cellophane:
@olamy merge conflicts resolved.

angry-cellophane:
@olamy thank you for merging the PR. 
Is it possible to release a new version of the plugin please?

PR #80, opened 2019-12-09
For JENKINS-58783: Only color stage start with global status

TobiX:
This makes it possible to see the status of individual stages.

Currently, the CSS for the global pipeline status overrides the status for each individual step. This change only applies the background color to the "start" stage instead of the complete row.

### DIFF
--- a/ui/src/main/js/view/templates/pipeline-staged.less
+++ b/ui/src/main/js/view/templates/pipeline-staged.less
@@ -54,10 +54,10 @@
     td.UNSTABLE .cell-color {
       background: rgba(230, 230, 9, .2)
     }
-    tr.UNSTABLE .cell-color {
+    tr.UNSTABLE td.stage-start .cell-color {
           background: rgba(230, 230, 9, .2)
     }
-    tr.FAILED .cell-color {
+    tr.FAILED td.stage-start .cell-color {
       background: rgba(255, 50, 0, .1)
     }
     tr.FAILED .FAILED .duration {


TobiX:
Here is a screenshot of the new style setup:

screenshot

Build 249 in this screenshot is correct, but looks strange: Every step was marked as successful, but somewhere in the build, something marked the whole build as failed...

michaelsembwever:
any eta on this? would make a "small" difference to many users, eg https://builds.apache.org/view/A-D/view/Cassandra/job/Cassandra-devbranch-pipeline/

dwnusbaum:
> Build 249 in this screenshot is correct, but looks strange: Every step was marked as successful, but somewhere in the build, something marked the whole build as failed...

I think this case is the reason that the current behavior is what it is, because the behavior with this PR was considered worse than showing the whole row as the overall result.

Generally this happens because the Pipeline uses the `step` or `wrap` steps to use a `SimpleBuilder` or `SimpleBuildWrapper`, which are also compatible with freestyle builds, which do not have steps, so they can only modify the build result. Sometimes it happens because of Pipeline steps that do the same thing (steps should not do this, they should use `WarningAction` to signify non-fatal problems). 

I don't think this totally fixes https://issues.jenkins-ci.org/browse/JENKINS-58783 unless it starts looking for `WarningAction` when computing stage statuses. With this PR as it is right now, a step like the current version of `junit` might have an unstable `WarningAction` and its stage should be shown as yellow, but it will be shown as green with this change. Similarly for `catchError` and `warnError`, and maybe also some steps for Warnings NG.

I don't think the PR as-is is problematic though, since no information is lost, it just shows strictly more information than was shown previously. It would just be even better if it looked at `WarningAction` too, and I don't think JENKINS-58783 should be closed until that happens.

olamy:
merged as it as temporary fix until better fix for JENKINS-58783

dwnusbaum:
Yeah, I don't mean to be negative: I think that this PR is better than the old logic, I just wanted to note that issues like build 249 in the screenshot cannot be resolved without looking at `WarningAction` (although even then, if the problem is a `SimpleBuilder` or `SimpleBuildWrapper`, the problem cannot currently be fixed).

PR #84, opened 2020-04-27
Avoid superfluous warnings

HrFlorianHoffmann:
The function `addExtensionPoints` logs warnings for stages' `<td>` elements that do not have a `data-stageId` attribute. In theory, all stages should have one. In practice, they only get their ID once they are started. Stages that are already pre-rendered as `NOT_EXECUTED` don't have one, though, and lead to lots of superfluous warnings in the concole logs.

### DIFF
--- a/ui/src/main/js/view/pipeline-staged.js
+++ b/ui/src/main/js/view/pipeline-staged.js
@@ -170,7 +170,9 @@ function addExtensionPoints(onElement, runGroupData) {
             var stageId = stage.attr('data-stageId');
 
             if (!stageId) {
-                console.warn('No "data-stageId" on stage.');
+                if (!stage.hasClass('NOT_EXECUTED')) {
+                    console.warn('No "data-stageId" on stage.');
+                }
                 return;
             }
 


ui/src/main/js/view/pipeline-staged.js
```diff
@@ -170,7 +170,9 @@ function addExtensionPoints(onElement, runGroupData) {
             var stageId = stage.attr('data-stageId');
 
             if (!stageId) {
-                console.warn('No "data-stageId" on stage.');
+                if (!stage.hasClass('NOT_EXECUTED')) {
```
alexanderrtaylor, line 173: should we not still warn about this though? Maybe just a lower log level?

We could easily console.info or something to make sure people understand that a stage has not been executed. I would think you still need to be told personally

HrFlorianHoffmann: This runs as part of the render method, every 5 seconds. A single build easily spawns thousands of these log messages. If we reduced only the log level, we would still be spamming the console.

The situation that a so-far-unprocessed stage doesn't have a `data-stageId` is nothing unexpected, that works just as designed. Such cases shouldn't be warnings at all. We should only warn people if a stage that _should_ have an ID, didn't get one. That's what I added the `.hasClass` for. (I'm a noob here, there may be better ways to do that. 😃)

I'm also not convinced the message is all that great. We basically want to tell developers why their extensions weren't added to a certain stage, right? Shouldn't the message then tell that, i.e. read something like "Didn't add extensions to stage 'Acceptance' because there was a technical error that it didn't receive a 'data-stageId'"?

alexanderrtaylor: What if you aggregated all of them into a single message with a list of all the stages it missed/skipped? So instead of just logging here, you would check if it is that hasClass and then add that stage name to a hash list or something and at the end of the build we can print it in the console

PR #85, opened 2020-04-28
Added the NOT_EXECUTED status

alexanderrtaylor:
There is a separate status of NOT_EXECUTED in the code but that is not accounted for in the javascript and so it was defaulting to "SUCCESS" which is pretty clearly not right.

Sorry no tests as I am not really sure how this would be tested to begin with but with local testing I can force a NOT_EXECUTED status with the code mentioned in the Jira and now it turns red

### DIFF
--- a/ui/src/main/js/view/templates/pipeline-staged.less
+++ b/ui/src/main/js/view/templates/pipeline-staged.less
@@ -411,6 +411,9 @@
   .FAILED .jobName .badge {
     background-color: @error-highlight;
   }
+  .NOT_EXECUTED .jobName .badge {
+      background-color: @error-highlight;
+   }
 
   .stage-cell.PENDING_START .stage-wrapper {
     border: none;


res0nance:
Is not executed the same as a failure? It might give the wrong impression

alexanderrtaylor:
If you look in the console logs, it marks the build "FAILED" and on disk that is the last failed build so I would think their technical status is the same

dwnusbaum:
> with the code mentioned in the Jira

Can you please link the Jira ticket in the PR description and add it to the PR title so we can see the context? Thanks!

alexanderrtaylor:
My bad @dwnusbaum forgot to put it in the title. https://issues.jenkins-ci.org/browse/JENKINS-61591 is the issue but you cant do a straight hpi:run on this code as you need to have a newer version of Jenkins and declarative plugins installed to reproduce (easy enough to change the pom.xml but I wanted to warn you)

PR #87, opened 2020-06-21
Add dark theme compatibility

timja:
Fixes https://github.com/jenkinsci/dark-theme-plugin/issues/85

When testing set jenkins.version to be 2.240 and plugin-pom version to be 4.3

<details>
<summary>Main screen - before changes</summary>

image

</details>

<details>
<summary>Logs pop up - before changes</summary>

image

</details>

<details>
<summary>Main screen - no theme</summary>

image

</details>

<details>
<summary>Logs pop up - no theme</summary>

image

</details>

<details>
<summary>Main screen - Dark theme</summary>

image

</details>

<details>
<summary>Logs pop up - Dark theme</summary>

image

</details>

cc @oleg-nenashev @olamy @fqueiruga

### DIFF
--- a/ui/src/main/js/view/widgets/dialog/style.less
+++ b/ui/src/main/js/view/widgets/dialog/style.less
@@ -10,6 +10,7 @@
     border-radius-topleft: @cbwf-border-radius;
     border-radius-topright: @cbwf-border-radius;
     background-color: @light-header-background-color;
+    // This variable may be renamed on Jenkins core in the future and would stop working
     background-color: var(--bigtable-header-color);
     padding: 7px 10px;
     font-weight: bold;
--- a/ui/src/main/js/view/widgets/dialog/style.less
+++ b/ui/src/main/js/view/widgets/dialog/style.less
@@ -1,13 +1,16 @@
 .cbwf-dialog {
   border-radius: @cbwf-border-radius;
   border: solid 1px @dark-border-color;
+  border: solid 1px var(--panel-border-color);
   background-color: @light-body-background-color;
+  background-color: var(--background);
   box-shadow:rgba(0,0,0,.25) 0 0 0 999em, rgba(0,0,0,.25) 0 0 0 1px, rgba(0,0,0,.33) 0 3px 10px;
 
   .header {
     border-radius-topleft: @cbwf-border-radius;
     border-radius-topright: @cbwf-border-radius;
     background-color: @light-header-background-color;
+    background-color: var(--bigtable-header-color);
     padding: 7px 10px;
     font-weight: bold;
     font-size: 1.2em;
--- a/ui/src/main/js/view/widgets/dialog/style.less
+++ b/ui/src/main/js/view/widgets/dialog/style.less
@@ -10,8 +10,7 @@
     border-radius-topleft: @cbwf-border-radius;
     border-radius-topright: @cbwf-border-radius;
     background-color: @light-header-background-color;
-    // This variable may be renamed on Jenkins core in the future and would stop working
-    background-color: var(--bigtable-header-color);
+    background-color: var(--bigtable-header-bg);
     padding: 7px 10px;
     font-weight: bold;
     font-size: 1.2em;
--- a/ui/src/main/less/stageview_adjunct.less
+++ b/ui/src/main/less/stageview_adjunct.less
@@ -8,6 +8,12 @@
 // Selectively import some twitter bootstrap styles.  Namespace them too.
 .cbwf-widget {
   @import "bootstrap";
+  
+  pre {
+    background-color: var(--background);
+    color: var(--text-color);
+    border-color: var(--panel-border-color);
+  }
 }
 
 div.fullscreen#main-panel {
--- a/ui/src/main/js/view/templates/pipeline-staged.less
+++ b/ui/src/main/js/view/templates/pipeline-staged.less
@@ -128,6 +128,7 @@
     .table-viewPort {
       display: inline-block;
       border: 1px solid #ccc;
+      border: 1px solid var(--panel-border-color);
       overflow-x: auto;
       overflow-y: hidden;
       box-shadow: rgba(0, 0, 0, .25) 1px 3px 5px;
@@ -325,6 +326,7 @@
 
   .stage-start-time, .stage-wrapper {
     border: solid 1px @light-border-color;
+    border: solid 1px var(--panel-border-color);
     border-radius: @cbwf-border-radius;
   }
 
@@ -388,7 +390,6 @@
         bottom: 15px;
         font-size: 60%;
         text-align: center;
-        .text-muted;
         opacity: 0.6;
       }
     }
--- a/ui/src/main/js/view/templates/pipeline-staged.less
+++ b/ui/src/main/js/view/templates/pipeline-staged.less
@@ -91,7 +91,6 @@
       }
       td {
         border: 1px solid rgba(0, 0, 0, .05);
-        text-shadow: #fff 0 1px 1px, #fff 0 1px 2px;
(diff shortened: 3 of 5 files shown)


ui/src/main/less/stageview_adjunct.less
```diff
@@ -8,6 +8,12 @@
 // Selectively import some twitter bootstrap styles.  Namespace them too.
 .cbwf-widget {
   @import "bootstrap";
+  
+  pre {
+    background-color: var(--background);
```
fqueiruga, line 13: Is this backwards-compatible?

ui/src/main/js/view/widgets/dialog/style.less
```diff
@@ -1,13 +1,16 @@
 .cbwf-dialog {
   border-radius: @cbwf-border-radius;
   border: solid 1px @dark-border-color;
+  border: solid 1px var(--panel-border-color);
```
fqueiruga, line 4: I'm a big reluctant on using this variables bc I think they may need to be refactored / renamed in the future.

ui/src/main/js/view/widgets/dialog/style.less
```diff
@@ -1,13 +1,16 @@
...
   background-color: @light-body-background-color;
+  background-color: var(--background);
   box-shadow:rgba(0,0,0,.25) 0 0 0 999em, rgba(0,0,0,.25) 0 0 0 1px, rgba(0,0,0,.33) 0 3px 10px;
 
   .header {
     border-radius-topleft: @cbwf-border-radius;
     border-radius-topright: @cbwf-border-radius;
     background-color: @light-header-background-color;
+    background-color: var(--bigtable-header-color);
```
fqueiruga, line 13: Ditto

ui/src/main/js/view/widgets/dialog/style.less
```diff
@@ -1,13 +1,16 @@
 .cbwf-dialog {
   border-radius: @cbwf-border-radius;
   border: solid 1px @dark-border-color;
+  border: solid 1px var(--panel-border-color);
   background-color: @light-body-background-color;
+  background-color: var(--background);
```
fqueiruga, line 6: Is this variable only found on the dark-theme? Shouldn't we only use variables defined on Jenkins core?

ui/src/main/less/stageview_adjunct.less
timja: yes, as it's overriding bootstrap styles, it will fallback to the previous styles otherwise

ui/src/main/js/view/widgets/dialog/style.less
timja: https://github.com/jenkinsci/jenkins/blob/master/war/src/main/less/abstracts/colors.less#L21

ui/src/main/js/view/widgets/dialog/style.less
timja: I think this one should be okay?

ui/src/main/js/view/widgets/dialog/style.less
fqueiruga: 👍

ui/src/main/js/view/widgets/dialog/style.less
timja: This one should probably be named better if it's to be used with more than just bigtable

ui/src/main/js/view/widgets/dialog/style.less
fqueiruga: This one is actually an example of one that's problematic. This one is used for the side panel and the realtime widgets border. Then there's `--pane-border-color` that is used as border on `.pane-frame`.

ui/src/main/js/view/widgets/dialog/style.less
fqueiruga: I'm probably going to be restyling the tables on Jenkins in a few weeks. Maybe we can discuss the CSS vars API for tables then.

ui/src/main/less/stageview_adjunct.less
fqueiruga: 👍

ui/src/main/js/view/widgets/dialog/style.less
timja: sure for now, I think we can add this, and dark-theme will support it, if it disappears from Jenkins it will just fallback to what was there previously

Once the API has been sorted we can adjust here and dark-theme can support both for a little bit

fqueiruga: Sounds like a plan!

fqueiruga: I would only then suggest leaving a comment indicating that this variable is brittle.

fqueiruga:
```suggestion
    // This variable may be renamed on Jenkins core in the future and stop working
    background-color: var(--bigtable-header-color);
```

timja: not sure if it's needed really, there's a fallback added.

timja:
@olamy would be great if you could take a look please

We've been running this in production from the incremental version for the last couple of weeks with no complaints, in my last commit I fixed up a couple of issues with borders and the paused text being unreadable

it's not perfect but it's far better than what's currently there.

timja:
@olamy would you be able to release this please? A couple of users have asked today https://github.com/jenkinsci/dark-theme-plugin/issues/85#issuecomment-669934404

Thanks!

ui/src/main/js/view/widgets/dialog/style.less
VirtualTim: I know I'm a bit late, but 
```css
background-color: @light-body-background-color;
background-color: var(--background);
```
Looks like a bug to me.

I also can't see where var(--background) is defined

olamy: @VirtualTim no worries mate. Maybe provide a PR fixing it?

VirtualTim: I don't really know enough about how this works for a PR. My guess is that the intention was to do `background-color: var(--background, @light-body-background-color);`, but I'm just guessing.

pappnase99:
Unfortunately this change now leads to the stage view being half-transparent and not really helpful when you're not using dark mode, like in screenshot.

Screenshot 2020-09-15 153511

The issue here is the statement `background-color: var(--background);`, if you disable it, it looks alright.

VirtualTim:
Yeah I had the same issue. Updating to latest Jenkins LTS fixed this for me.
The issue (I believe) is doing this
```css
background-color: @light-body-background-color;
background-color: var(--background);
```
where it should be this:
```css
background-color: var(--background, @light-body-background-color);
```
But I wasn't confident enough in this fix to submit a PR.

timja:
That’s not the issue, (not sure where it is), the double property syntax is used because IE 11 doesn’t support css variables.

I tested this when I added in the last fix here and was unable to reproduce.

I would recommend upgrading to the latest release of Jenkins since it’s fixed there

VirtualTim:
@timja it's easy to test, just replace 
```css
background-color: @light-body-background-color;
background-color: var(--background);
```
with
```css
background-color: @light-body-background-color;
background-color: var(--background-var-doesnt-exist);
```

So this only occurs when `--background` isn't defined, which I think was briefly the case in the LTS version.

At least that explains the double property syntax.

timja:
Ah sure it needs to be this then:

```
background-color: @light-body-background-color;
background-color: var(--background, @light-body-background-color);
```

PR #88, opened 2020-07-16
Fix Inconsistent Parallel Stage Order

lsouchard:
This solution is offered to resolve the JIRA [JENKINS-63115](https://issues.jenkins-ci.org/browse/JENKINS-63115) that I also created to detail the issue. This solution borrows from the PR #63 from @Nuh which have become stale.

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/RunExt.java
@@ -40,6 +40,7 @@ import org.jenkinsci.plugins.workflow.support.steps.input.InputStepExecution;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.concurrent.TimeoutException;
 
@@ -316,7 +317,9 @@ public class RunExt {
         if (execution != null) {
             ChunkVisitor visitor = new ChunkVisitor(run);
             ForkScanner.visitSimpleChunks(execution.getCurrentHeads(), visitor, new StageChunkFinder());
-            runExt.setStages(new ArrayList<StageNodeExt>(visitor.stages));
+            final List<StageNodeExt> visitorStages = new ArrayList<>(visitor.stages);
+            visitorStages.sort(RunExt::compareStageNodeExt);
+            runExt.setStages(visitorStages);
         }
 
         long currentTimeMillis = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


anthonyemerson:
is it possible this would also fix https://issues.jenkins-ci.org/browse/JENKINS-62366 ? I suspect 62366 happens because the parallel stages change the run order which Jenkins registers as a new config and so the old stage history is wiped while running

lsouchard:
Yes it should, I didn't see this JIRA and created my own but the behavior described is almost the same. For some reason with parallel stages `ForkScanner.visitSimpleChunks` doesn't visit the nodes of a run in the same order while running and once done. Unfortunately this is too deep in Jenkins API to fix it myself. That's why I dealt with it on this plugin's API directly after visiting and collecting the stage nodes. To be clear the behavior seen in the screenshots of both JIRAs is resulting from displaying lines of runs starting from the most recent one and adding older runs line by line that are only allowed to have more subsequent stages than the one above and stopping as soon as a run break that rule.

anthonyemerson:
@svanoort Any chance we can get this prioritized to go in? Seems a simple fix for a major quality of life issue

jquinc30:
@olamy is still active on the project, any chance to have a quick look at this pr as changes are pretty small ?

olamy:
would be good to have some tests for this to avoid regression in the future.

anthonyemerson:
@lsouchard you up for adding testing around this?

lsouchard:
@anthonyemerson Yes I will. I am in the middle of something right now but I will add some probably later this week.

anthonyemerson:
Anyone else super excited!, Thanks @lsouchard This is going to make my life sooooo much easier

lsouchard:
Here is a unit test to check it. I would have preferred to fully test the sorting of the stage nodes in RunExt#createNew right after the ChunkVisitor is done collecting the stages and comparing them. However, the static nature of all these methods prevents me from mocking any of these. The only solution would have been to create a mock of WorkflowRun for the ChunkVisitor to collect stages from, but the way this visitor do so is to obscure for me to build the mock I need on an object so massive. About my test, I am sorry to use sleep in the script but I haven't found any way to share a lock between the script and the test (groovy's sandbox preventing from accessing any singleton). So I had to allow the script to sleep for 2 seconds, It seems plenty enough to query the API and gather the data needed to test the order of the stages while the job is running. If anyone has any advice on how to improve on those issues, I would gladly implement those changes.

olamy:
@lsouchard thanks a lot that's a good test! I perfectly understand how it's a pain :)

agabrys:
Hello 🙂 The fix doesn't solve the issue (I tested the latest release [2.17](https://github.com/jenkinsci/pipeline-stage-view-plugin/releases/tag/pipeline-stage-view-2.17)). I described the problem here: [JENKINS-63115](https://issues.jenkins-ci.org/browse/JENKINS-63115?focusedCommentId=399364&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-399364).

PR #105, opened 2020-09-08
Fix plugin site GitHub link

timja:
Should fix https://github.com/jenkins-infra/plugin-site/issues/350


halkeye (approved):

olamy:
why do we need that? those poms should inherit values from parent pom.

halkeye:
They absolutely should but the pom files have the plugin name appended. It might be an issue with the pom plugin?

olamy:
haha I know. Maven was build on legacy svn so the `artifactId` value is added to the parent pom value which makes the result wrong :)

olamy (approved):

PR #41, opened 2017-05-24
Maximize popup window size for using screen estate

ssbarnea:
[JENKINS-44470](https://issues.jenkins-ci.org/browse/JENKINS-44470)

PR #107, opened 2020-09-16
Remove JS calculation of width, just use css.

VirtualTim:
Also set the width to (100%-10px), so this can make better use of the screen real-estate.
Should help address [JENKINS-44470](https://issues.jenkins-ci.org/browse/JENKINS-44470), but could have some more improvements. Similar idea to #41. 

**Screenshots**
<details><summary>Before</summary>

One

Two

</details>

<details><summary>After</summary>

One

Two

</details>

`calc` is [supported from IE9](https://developer.mozilla.org/en-US/docs/Web/CSS/calc#Browser_compatibility), so hopefully that's fine. But being in the .less may mean that this works in earlier IE versions, but I haven't tested it.

@ssbarnea, @svanoort are you interested in taking a look?

### DIFF
--- a/ui/src/main/js/view/stage-logs.js
+++ b/ui/src/main/js/view/stage-logs.js
@@ -29,9 +29,7 @@ exports.render = function (stageDescription, onElement) {
     var $ = jqProxy.getJQuery();
     var theWindow = require('window-handle').getWindow();
     var stageLogsDom = templates.apply('stage-logs', stageDescription);
-    var winWidth = $(theWindow).width();
     var winHeight = $(theWindow).height();
-    var dialogWidth = Math.min(800, (winWidth * 0.7));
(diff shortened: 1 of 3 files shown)


timja:
@olamy

VirtualTim:
I don't know if this is something people are interested in, but as it is there's lots of wasted screen space and this seems like a simple way to address that.

timja:
@VirtuaTim there’s only one person who merged here in genera, there’s usually a bit of a wait time in this repo

ssbarnea (approved):

timja:
Might have caused https://issues.jenkins-ci.org/browse/JENKINS-63892

VirtualTim:
Yeah, I believe it did. I've opened #118 to try and fix it.

PR #108, opened 2020-09-19
Fix counting maxRunsPerJob

ngg:
JobExt.create() used to return maxRunsPerJob + 1 runs instead of
maxRunsPerJob. Moving the check before the addition makes it work if "0"
is set as well.

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/JobExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/JobExt.java
@@ -40,9 +40,14 @@ import java.util.List;
 public class JobExt {
 
     /**
-     * Max number of runs per page. Pagination not yet supported.
+     * Name of the system property to set the max number of runs per page. Pagination not yet supported.
      */
-    public static final int MAX_RUNS_PER_JOB = Integer.getInteger(JobExt.class.getName()+".maxRunsPerJob", 10);
+    public static final String MAX_RUNS_PER_JOB_PROPERTY_NAME = JobExt.class.getName()+".maxRunsPerJob";
+
+    /**
+     * Default value of the max number of runs per page.
+     */
+    public static final int MAX_RUNS_PER_JOB_DEFAULT = 10;
 
     private JobLinks _links;
     private String name;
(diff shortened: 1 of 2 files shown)


R-Rudolf:
This PR would enable to decrease the maxRunsPerJob to zero, thus disabling the stage view completely.
It is a use case, required by some people, see: https://stackoverflow.com/questions/52438258/how-to-disable-jenkins-pipeline-stage-view-plugin
Related ticket: https://issues.jenkins-ci.org/browse/JENKINS-63614

olamy:
sounds good but having a test would be even better to avoid future regressions

ngg:
I'll try to add some tests, but need some advice.
I think the best place for these tests would be in JobAndRunAPITest.java, please tell me if I'm wrong.
The problem is that this parameter is not easily changable from tests as it is a static integer:
```java
public static final int MAX_RUNS_PER_JOB = Integer.getInteger(JobExt.class.getName()+".maxRunsPerJob", 10);
```
I couldn't find any tests which sets any of the system properties, so I cannot just use the same solution.
Some ideas how to solve this:
- Move the `Integer.getInteger()` call into `JobExt.create()` where it is used, so tests can use `System.setProperty` to override the value before calling the API.
- Add a new parameter to `JobExt.create()` and possibly a new `@QueryParameter` to `JobAPI.doRuns()` which can override the default value
- Create separate test classes for these tests and use some kind of `@BeforeClass`/`@AfterClass` magic to set and restore the property before evaluating the static field. (I'm not sure if this would actually work)

@olamy What do you think which would be the proper way to test it? Or do you have some other idea?

olamy:
@ngg good question mate!
I guess the logic with a separate test class and `@BeforeClass/@AfterClass` should work

ngg:
@olamy `@BeforeClass`  unfortunately does not work, because the static `MAX_RUNS_PER_JOB` is only evaluated once. So whichever test runs first that uses the `JobExt` class sets it to some value (which really can be overriden with `@BeforeClass`, but later tests use that same value because the property is only queried once.
I'll update my PR and go with my first idea: moving the `getInteger()` call where it is used, because that does not cause an unnecessary API change and also has the nice side-effect that this property can be changed without Jenkins restarts.

PR #116, opened 2020-10-08
Fix calc not being escaped

VirtualTim:
Fix for issue introduced in #107.
I didn't realise I had to escape the calc. Without escaping less will replace it with `calc(90%)`, which is not the same thing.

### DIFF
--- a/ui/src/main/less/stageview_adjunct.less
+++ b/ui/src/main/less/stageview_adjunct.less
@@ -14,7 +14,7 @@
     color: var(--text-color);
     border-color: var(--panel-border-color);
   }
-  width: calc(100% - 10px);
+  width: calc(~'100% - 10px');
 }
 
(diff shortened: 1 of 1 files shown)


VirtualTim:
Thanks @olamy. Sorry I didn't catch that originally.

snago:
@VirtualTim Any idea why this change wouldn't take effect in the 2.17 release?

See https://issues.jenkins-ci.org/browse/JENKINS-63892

TobiX:
@snago It's there, but somehow having the `width:` on `.cbwf-widget` at all prevents the input popup from showing up in the right location...

VirtualTim:
Crap, I've never used the `input` stage, so I missed that.
I'll do some testing, but if anyone knows a fix let me know (or submit a PR).

PR #118, opened 2020-10-13
Move width calculation from .cbwf-widget to .cbwf-widget.cbwf-modal

VirtualTim:
This should mean it only affects the larger popup, not the smaller one.

I believe this will fix JENKINS-63892, JENKINS-63914 and JENKINS-63901.

@snago, @TobiX are either of you able to test this with your workflows? I can now see the popup with the `input` stage, but I may have missed something else.

### DIFF
--- a/ui/src/main/less/stageview_adjunct.less
+++ b/ui/src/main/less/stageview_adjunct.less
@@ -14,6 +14,9 @@
     color: var(--text-color);
     border-color: var(--panel-border-color);
   }
+}
+
+.cbwf-widget.cbwf-modal {
   width: calc(~'100% - 10px');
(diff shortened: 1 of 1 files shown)


VirtualTim:
Build failed because the build system ran out of space: `There is not enough space on the disk.`
@olamy?

TobiX:
@VirtualTim I could build and test this on a test system and it works as intended (at least for me)

bungeeskimmer:
Works for me. Thanks!

olamy (approved):

timja:
could this be released please there's about 4 jira issues opened for it

PR #119, opened 2020-10-20
Allow input popovers to be bigger

TobiX:
@VirtualTim inspired me to look at some other QoL improvements for this plugin, this time in the input popup. When adding an input with many options, the popup overflows quite early.

Before:
grafik

After:
grafik

Still fine for small popups:
grafik

There are some other nits I have with this popup, but I haven't found good solutions for them yet:
- Closes too fast when mouse leaves the popup
- Opens too the left of the associated stage, but changing that may push the popup out of the browser window to the right if it is used in later stages...


TobiX:
@olamy Could you take a look?

PR #120, opened 2020-10-20
Keep input popovers open for longer

TobiX:
In #119 I mentioned that I think the input popovers close too fast and now I had some time to search for the cause.

This splits the "open" and "close" delays (were mixed in `inoutDelay` before) - The open delay is now a fixed 250ms (was 500ms before) and the close delay stays at 500ms by default, but the input popover overrides this to 3s - this makes it much more forgiving to select something in the input popover without it accidentally closing on the user.

Maybe 3s is even too long?


olamy:
ok for 3s. let see if users complains :)

PR #193, opened 2022-01-21
Fix misalignment of build icons since 2.329

timja:
See https://github.com/jenkinsci/jenkins/pull/6162#discussion_r782846360

This code has been here since open sourcing and was never touched.

As far as I can tell it's dead code, I couldn't see any references to it and can't think where it could be used as part of this plugin


NotMyFault (changes requested):
Need to remove the import from `/icons` in `stageview_adjunct` as well.

NotMyFault (approved):

MarkEWaite:
I've included the incremental build of the master branch in my [set of plugins](https://github.com/MarkEWaite/docker-lfs/blob/e900b7f31b7b2270d9578dca4f0d15579ed9cf96/plugins.txt#L140) and have confirmed that the issue is resolved.

The sooner we get a release of the plugin, the fewer issue reports we'll see for the misalignment.

PR #199, opened 2022-01-27
Fix link for build details

HrFlorianHoffmann:
Not entirely sure, but I think this is responsible for https://issues.jenkins.io/browse/JENKINS-52385. The links behind the numbers (e.g. "#5508") at the left side of the Full Stage View are incorrect: They all have a sub-path ".../workflow-stage/..." that shouldn't be there. The links end up nowhere, with an HTTP 404 from stapler.


felipecrs:
I can confirm the bug, but I have not tested this PR itself.

felipecrs:
I tested it, and unfortunately it does not fix the issue for me. Too bad, because it actually makes a big difference with `org.jenkinsci.pipeline.stageview.disabledOnMainJobPage=true`.


https://github.com/user-attachments/assets/485ee419-87a9-4323-9e9d-4593bf87e947

PR #201, opened 2022-02-05
Update plugin POM and BOM

basil:


basil:
@jenkinsci/pipeline-stage-view-plugin-developers A release of this would facilitate Java 17 PCT testing.

PR #205, opened 2022-02-21
Revert text-shadow change until dark mode regression is fixed

VirtualTim:
Change introduced in #146 makes times near-unreadable in dark mode. Roll back the change until the regression is resolved.
@timbrown5 @tdspencer3
See https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/146#issuecomment-1029704850

- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue


olamy:
thanks!
better to revert til we find a better solution.

PR #243, opened 2022-08-25
Fix isStageNode function

zuzuviewer:
Fix isStageNode function

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
@@ -30,6 +30,8 @@ import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.actions.ThreadNameAction;
 import org.jenkinsci.plugins.workflow.graph.AtomNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.graph.StepNode;
+import org.jenkinsci.plugins.workflow.support.steps.StageStep;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
```diff
@@ -58,7 +60,18 @@ public void setStageFlowNodes(List<AtomFlowNodeExt> stageFlowNodes) {
     }
 
     public static boolean isStageNode(FlowNode node) {
-        return (node.getAction(LabelAction.class) != null && node.getAction(ThreadNameAction.class) == null);
+        final boolean stageNode = (node.getAction(LabelAction.class) != null && node.getAction(ThreadNameAction.class) == null);
```
olamy, line 63: no need of final.

olamy (changes requested):

rest-api/src/main/java/com/cloudbees/workflow/rest/external/StageNodeExt.java
zuzuviewer: > no need of final.

Why？

olamy: well I could ask why adding final for a local variable?
frankly it doesn't change anything nor adding any value except adding more (useless?) keyword :) 
if you think it's very important and improve performance or anything else I will not complain.
but I don't see the real added value.
well maybe it's nit picking :)

zuzuviewer: Ok, I think it's a good way to tell other people this is unmodifiable variable, and don't try to change it's value, and this is my habit.
I agress with you it is not very important in here , but it's not a bad thing.
So if you still think it's no need, I will remove it;

olamy: > Ok, I think it's a good way to tell other people this is unmodifiable variable, and don't try to change it's value, and this is my habit. I agress with you it is not very important in here , but it's not a bad thing. So if you still think it's no need, I will remove it;

why not but why here?
it's a local variable of a very small method use only once in the method body. 
So you probably don't even need to use a local variable.

I cannot find any unit test for this change.
Is it something you could add please?

zuzuviewer: I add a test for it

PR #316, opened 2023-10-02
Add Jdk21 build

gounthar:
Test with Java 21
==

Java 21 was released Sep 19, 2023. We want to announce full support for Java 21 in early October and would like the most used plugins to be compiled and tested with Java 21.

The acceptance test harness and plugin bill of materials tests are already passing with Java 21. This is a further step to improve plugin readiness for use with Java 21 and for development with Java 21.

The change intentionally tests only two Java configurations, Java 17 and Java 21 because we believe that the risk of a regression that only affects Java 11 is shallow. We generate Java 11 byte code with the Java 17 and the Java 21 builds, so we're already testing Java 11 byte code.

Testing done
===

Confirmed tests pass with Java 21


olamy:
builds looks to be stucked to get windows jdk11 node

gounthar:
Yes, it's been stuck for the last 17 hours.

Jenkinsfile
```diff
@@ -1,5 +1,7 @@
-buildPlugin(useContainerAgent: false, configurations: [
-  [ platform: "linux", jdk: "11" ],
-  [ platform: "windows", jdk: "11" ],
-  [ platform: "linux", jdk: "17" ]
+buildPlugin(
+  useContainerAgent: true,
+  configurations: [
+    [platform: "linux", jdk: "11"],
+    [platform: 'linux', jdk: 21],
+    [platform: 'windows', jdk: 17],
```
olamy, line 6: Probably no need of 17.
Per my understanding Jenkins will jump directly from 11 to 21.
So 17 seems to be a no need for testing?
I would rather prefer having windows 11 and/or 21.

gounthar: Starting soon, the weekly Jenkins releases will include an administrative monitor to notify users that Jenkins will no longer support Java 11 after September 30, 2024.

The official documentation is currently being updated to support JDK17. Additionally, starting this week, the default Docker images will use JDK17. JDK17 is now the preferred version for Jenkins until we eventually transition to JDK21.

For testing purposes, it's important to check the compatibility of your plugin with both JDK17 and JDK21. Regarding JDK11, the choice to continue using it is up to you. 🤗

PR #349, opened 2024-09-27
Jenkins 73817 Add a Configuration Parameter to Control Log Output Format (Text vs. HTML)

kun303-cloud:
Currently, our logging system does not provide a direct way to switch between pure text and HTML-formatted log outputs. This limitation can be restrictive in scenarios where logs need to be rendered directly in web interfaces or where specific formatting requirements (e.g., for highlighting, linking, or styling) are desired.

To enhance the flexibility and usability of our logging capabilities, I propose the addition of a new configuration parameter that would allow users to specify the desired output format for logs. This parameter could be easily integrated into our existing logging framework, allowing for seamless transitions between text and HTML modes as needed.

Example:

http://{ip_host}/job/{optionalFolderPath}/job/{job_name}/{build_number}/execution/node/{nodeId}/wfapi/log?text=true

### DIFF
--- a/rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/FlowNodeAPI.java
+++ b/rest-api/src/main/java/com/cloudbees/workflow/rest/endpoints/FlowNodeAPI.java
@@ -30,6 +30,7 @@ import com.cloudbees.workflow.util.ModelUtil;
 import com.cloudbees.workflow.util.ServeJson;
 import hudson.Extension;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.kohsuke.stapler.QueryParameter;
 
 import java.io.IOException;
 
@@ -59,7 +60,8 @@ public class FlowNodeAPI extends AbstractFlowNodeActionHandler {
     }
 
     @ServeJson
-    public Object doLog() {
-        return Log.get(getNode());
+    public Object doLog(@QueryParameter boolean text) {
+        System.out.println("doLog:text:"+text);
+        return text?Log.getText(getNode()):Log.getHtmlText(getNode());
     }
 }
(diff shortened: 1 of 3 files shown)


jglick (changes requested):
The HTTP endpoints in the REST plugin in this repo are intended solely for use by the stage view GUI, so unless there is some fix visible in the GUI that this allows, it should not be done.

kun303-cloud:
> The HTTP endpoints in the REST plugin in this repo are intended solely for use by the stage view GUI, so unless there is some fix visible in the GUI that this allows, it should not be done.

You’re absolutely right, but this REST plugin has already become a standard part of Jenkins. In many integration scenarios, the use case is not based on a web GUI. Considering the above scenarios, that's why I modified these endpoints, so I hope you can take similar scenarios into account.

jglick:
> this REST plugin has already become a standard part of Jenkins

It is not standard and is not supported. If you have some other application for which you need a REST API rendering stage/step metadata, logs, etc., you will need to write your own REST API plugin.

kun303-cloud:
> > this REST plugin has already become a standard part of Jenkins
> 
> It is not standard and is not supported. If you have some other application for which you need a REST API rendering stage/step metadata, logs, etc., you will need to write your own REST API plugin.

But I don't think that's reasonable. I integrated Jenkins and used the officially recommended java-rest library. This library uses this and the plugin API. Of course, BlueOcean also supports plain text log operations. But the whole integration process was frustrating, and not only did I need to install an extra plugin, but I also had to write and operate the Blue Ocean API myself.

kun303-cloud:
> The REST APIs from `blueocean-plugin` were designed for _that_ visualization and also not supported for any other purpose.

[jenkins-rest](https://www.jenkins.io/projects/gsoc/2020/project-ideas/jenkins-rest-plugin/)
The methods runHistory, workflow, pipelineNode, and pipelineNodeLog all utilize the APIs from the pipeline-stage-view-plugin. Therefore, I would like to write a method that is compatible with them. Currently, there is no relevant compatibility method available, but it's fine; I will improve and complete the content I need later. Thank you for your reply.

PR #369, opened 2025-02-13
Add the missing forward slash to run relative url

gbhat618:
A Run object URL in Jenkins should end with a forward slash. Pipeline stage view run object is not attaching the forward slash - Jenkins has a mechanism to automatically redirect from the non forward slash to with forward slash so it is not a visible problem to the user.

<details><summary>Redirection happening in Jenkins core</summary>

However as we can see it still wastes a network call.

|Description|Screenshot|
|---|---|
|Notice a first request returned with 302, and then the proper request comes with 200|image|
</details>

This issue was found via the CBCI high availability controller we noticed it was flagged for builds currently in execution (these kind of cases are deliberately caught so to fix in the upstream source).

Fixing it in here would be the correct fix, as it also reduces one network call to the server.

**Testing**
* fixed the automated test failures
* manual testing
<details><summary>Screenshots before and after</summary>

|Description|Before Screenshot|After Screenshot|
|----|----|----|
|Hoverover the run id to see the resolved full URL|image|image|
|Browser inspect show the relative URL correctly fixed|image|image|
|Opening a run page does have redirects anymore - see single call|image|image|

</details>

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/ui/src/main/js/view/templates/pipeline-staged.hbs
+++ b/ui/src/main/js/view/templates/pipeline-staged.hbs
@@ -43,7 +43,7 @@
         <td class="stage-start">
           <div class="cell-color">
             <div class="cell-box">
-              <div class="jobName"><span class="badge"><a href="{{#if this.id}}{{this.id}}/{{else}}{{/if}}">{{this.name}}</a></span></div>
+              <div class="jobName"><span class="badge"><a href="{{#if this.id}}{{this.id}}/{{/if}}">{{this.name}}</a></span></div>
               <div class="stage-start-box">
                 <div class="stage-start-time">
                   <div class="date">{{formatDate this.startTimeMillis 'month'}} {{formatDate this.startTimeMillis 'dom'}}</div>
--- a/ui/src/main/js/view/templates/pipeline-staged.hbs
+++ b/ui/src/main/js/view/templates/pipeline-staged.hbs
@@ -43,7 +43,7 @@
         <td class="stage-start">
           <div class="cell-color">
             <div class="cell-box">
-              <div class="jobName"><span class="badge"><a href="{{this.id}}">{{this.name}}</a></span></div>
+              <div class="jobName"><span class="badge"><a href="{{this.id}}/">{{this.name}}</a></span></div>
               <div class="stage-start-box">
                 <div class="stage-start-time">
                   <div class="date">{{formatDate this.startTimeMillis 'month'}} {{formatDate this.startTimeMillis 'dom'}}</div>
--- a/ui/src/main/js/view/templates/pipeline-staged.hbs
+++ b/ui/src/main/js/view/templates/pipeline-staged.hbs
@@ -43,7 +43,7 @@
         <td class="stage-start">
           <div class="cell-color">
             <div class="cell-box">
-              <div class="jobName"><span class="badge"><a href="{{this.id}}/">{{this.name}}</a></span></div>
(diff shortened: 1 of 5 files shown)


gbhat618:
(looking at the test failure)

ui/src/main/js/view/templates/pipeline-staged.hbs
```diff
@@ -43,7 +43,7 @@
         <td class="stage-start">
           <div class="cell-color">
             <div class="cell-box">
-              <div class="jobName"><span class="badge"><a href="{{this.id}}">{{this.name}}</a></span></div>
+              <div class="jobName"><span class="badge"><a href="{{#if this.id}}{{this.id}}/{{else}}{{/if}}">{{this.name}}</a></span></div>
```
gbhat618, line 46: instead of just doing `"{{this.id}}/"` I have added the `#if...` because in the test fixtures there are blanks expected;
* https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/296c306cdbcc734c407790be6f25b504d05218e4/ui/src/test/resources/view/pipeline_staged/render/expected.html#L50-L50
* https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/296c306cdbcc734c407790be6f25b504d05218e4/ui/src/test/resources/view/pipeline_staged/render/expected.html#L108-L108
* https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/296c306cdbcc734c407790be6f25b504d05218e4/ui/src/test/resources/view/pipeline_staged/render/expected.html#L163-L163

I am not sure if it is correct or not; how the runId can be null or empty 🤔 while the run name is there..

I thought I would simply update the blank href to corresponding runIds, and also fix the fixture json files;
but then I wasn't sure if it was deliberately made like this..

ui/src/main/js/view/templates/pipeline-staged.hbs
```diff
@@ -43,7 +43,7 @@
         <td class="stage-start">
           <div class="cell-color">
             <div class="cell-box">
-              <div class="jobName"><span class="badge"><a href="{{this.id}}">{{this.name}}</a></span></div>
+              <div class="jobName"><span class="badge"><a href="{{#if this.id}}{{this.id}}/{{/if}}">{{this.name}}</a></span></div>
```
gbhat618, line 46: instead of just doing `"{{this.id}}/"` I have added the `#if...` because in the test fixtures there are blanks expected;
* https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/296c306cdbcc734c407790be6f25b504d05218e4/ui/src/test/resources/view/pipeline_staged/render/expected.html#L50-L50
* https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/296c306cdbcc734c407790be6f25b504d05218e4/ui/src/test/resources/view/pipeline_staged/render/expected.html#L108-L108
* https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/296c306cdbcc734c407790be6f25b504d05218e4/ui/src/test/resources/view/pipeline_staged/render/expected.html#L163-L163

I am not sure if it is correct or not; how the runId can be null or empty 🤔 while the run name is there..

I thought I would simply update the blank href to corresponding runIds, and also fix the fixture json files;
but then I wasn't sure if it was deliberately made like this..

jglick (approved):

jglick:
@olamy please release, or better yet https://www.jenkins.io/doc/developer/publishing/releasing-cd/

PR #398, opened 2025-08-10
[JENKINS-35656] Use Jenkins user timezone or browser timezone for stage view

apuig:
Ensure that  stage start **date and time** displays in the UI use the Jenkins user's preferred timezone if set, or the browser's timezone as a fallback. The system timezone is **never** used for display, aligning the stage view plugin with the job build list widget, [HistoryWidget](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/widgets/HistoryWidget/entry.jelly), and reverting the behavior introduced in  [#236](https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/236)

**With this change, the system timezone is no longer used for display. The precedence is:**
1. Jenkins user preferences - `hudson.model.TimeZoneProperty` (via jelly tag `h.userTimeZone`)
2. Browser preferences (via `Intl.DateTimeFormat().resolvedOptions().timeZone`)

:speaking_head:  :information_source:  User locale and the format of the date/time (e.g., using 12h with or without AM/PM) are out of the scope of this PR, as adapting to those would require additional CSS changes to avoid breaking the render. `index.js` **still** contains this _hardcoded_ values : `month: 'short', day: '2-digit', hour: '2-digit', minute: '2-digit', hour12: false`.
**This PR address the data but reuse the same presentation**.

### :mega: Testing Notes 👻🚫 
- ~~**Currently, all the HTML tests are disabled pending a fix for PhantomJS.**~~
- To ensure the logic in `index.js` is executed, a standard Selenium driver is now used in the test. If there are better approaches (e.g., using ATH), suggestions are welcome.
  - **All the disabled `PhantomJS` tests are now migrated to `WebDriver` chrome**
- `WorkflowStageViewActionTest` has been updated to use HtmlUnit. However, HtmlUnit does not support `toLocaleDateString`, which required extra effort and a hell of debugging :rage3: .

###  Testing done
Apart of the introduced ITs I manually verified this using this steps:
* mvn hpi:run
* create a simple pipeline with stages and execute it
image

* create a user and use the script console to update 
```groovy
def user = hudson.model.User.getById("userInAustralia", false)
user.addProperty(new hudson.model.TimeZoneProperty("Australia/Sydney"))
user.save()
```
* login with this user
image

> [!NOTE]  
> 8:30 PM vs 20:30

* **before the fix**
image

### REFERENCES
- See [JENKINS-35656](https://issues.jenkins.io/browse/JENKINS-35656)
- See [JENKINS-70175](https://issues.jenkins.io/browse/JENKINS-70175) 
- Reference: [#236](https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/236)

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -154,11 +154,11 @@ So far we've ben writing two types of tests for the UI components:
 
 1. JavaScript based unit tests for testing the JavaScript components.  These tests are more lightweight and should run more quickly (no Jenkins instance, no browser env).
 Please see the [JavaScript Testing][js_testing] docs.
-1. More coarse grained Java based function tests using a combination of [`JenkinsRule`][JenkinsRule] and [PhantomJS] (headless browser)
-e.g. [`WorkflowJobActionTest`][WorkflowJobActionTest].  Extend [`AbstractPhantomJSTest`][AbstractPhantomJSTest] for these types of test.
-[`JenkinsRule`][JenkinsRule] allows us to create pipelines in a Jenkins instance embedded in the test, while [PhantomJS] allows us to
+1. More coarse grained Java based function tests using a combination of [`JenkinsRule`][JenkinsRule] and [WebDriver] (headless browser)
+e.g. [`WorkflowJobActionTest`][WorkflowJobActionTest].  Extend [`AbstractWebDriverTest`][AbstractWebDriverTest] for these types of test.
+[`JenkinsRule`][JenkinsRule] allows us to create pipelines in a Jenkins instance embedded in the test, while [WebDriver] allows us to
 request the relevant pages to be tested from that Jenkins instance.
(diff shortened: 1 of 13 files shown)


MarkEWaite:
@apuig the 2.40 release of this plugin fails a test in the Jenkins plugin BOM and fails the test when run in this repository with the command:

```
mvn -Djenkins.version=2.541.3 -Dtest=com.cloudbees.workflow.ui.view.WorkflowStageViewActionTest#testTimezoneRespectedInStageStartTime clean verify
```

@darinpope I'll submit a pull request to revert the 2.40 update from the plugin BOM so that we don't block the weekly release and don't block the inclusion of the update to jackson3-api in the plugin BOM.

@olamy, it would be a great help if the jackson2-api could be released with the update in pull request:

* https://github.com/jenkinsci/jackson2-api-plugin/pull/336

